PR #34, opened 2016-01-16
Adding support for Pogostick

antonym:
Pogostick - Offline Windows Password and Registry Editor
http://pogostick.net/~pnh/ntpasswd


chris18890:
worked like a charm! ;D thanks!

PR #35, opened 2016-01-17
Add support for AVG; closes #24

### LINKED ISSUE #24: AV option
I think a really useful option is an offline AV scanner.

Personally, I use the [AVG Rescue CD](https://share.avg.com/arl/) via PXE but unsure if it can be used in terms of licensing issues, but any other suggestions are also welcome!

chris18890:
Add support for AVG; closes #24


antonym:
LGTM, thanks!

PR #51, opened 2016-01-20
Download cdXY.iso instead of installXY.iso

viq:
Since the iso image is not available from the installer, there's no point in downloading 200+ MB installXY.iso when we can get the same install environment from <10MB cdXY.iso


antonym:
Thanks, didn't know about that one, looks like it works great!

PR #52, opened 2016-01-20
Adding OpenBSD Snapshots

antonym:


viq:
Thank you. This will need to be changed every few months though. Maybe it could be grabbed automatically, eg from http://ftp.openbsd.org/pub/OpenBSD/snapshots/amd64/index.txt ? Or is that too much complication for something this simple?

antonym:
My guess is it will change when the release occurs, which is when we'll need to update the menu anyways.

viq:
I think you'll get 6.0-beta somewhere around March/April, whereas release will be in May, so edits will need to be made ~4 times a year.

antonym:
Ok, we can revisit when that happens, I'll keep that in mind.  If it breaks, just let me know and we can work on it.  I've had some distributions change things up completely within a new major release too so they always keep me on my toes.

viq:
OpenBSD doesn't seem to be eager to make huge changes, so it most likely will be a simple case of s/59/60/ and so on, I just thought it could be nice to not even have to do that ;)

I'll try to remember, right now I'm in the "ooh, shiny, what does it do?" stage ;)

PR #97, opened 2016-04-25
Add MirBSD (MirOS BSD-current/i386 install/rescue boot)

mirabilos:
without image signatures, for now (until someone tells me how to add them)

Note: chainloading is broken, it does load menu.ipxe from the github branch, but as soon as I enter the BSD menu I get the original from your boot server; if I go to the iPXE shell and then run `chain --autofree 
https://raw.githubusercontent.com/MirBSD/netboot.xyz/master/src/bsd.ipxe` it can be tested.

### DIFF
--- a/src/bsd.ipxe
+++ b/src/bsd.ipxe
@@ -11,6 +11,7 @@ iseq ${arch} x86_64 && set arch_a amd64 || set arch_a ${arch}
 item --gap BSD Based Operating Systems
 item freebsd ${space} FreeBSD
 item openbsd ${space} OpenBSD
+item mirbsd ${space} MirBSD (i386 only)
 
 # Options
 item --gap Options:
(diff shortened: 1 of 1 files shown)


antonym:
Thanks, I gave this a try and it appears to work well.  I'll go ahead and add it.  It also appears it's referencing the now defunct netboot.me if you want to change it.

If you guys get releases staged out, we can always add the checksums. (just difficult for rolling right now, haven't figured out the easiest low-maintenance way for that yet)

For the future as an FYI, when testing a remote branch from github, all you have to do is disable the signature checks, since it's attempting to use the original code signature and the modified file on github doesn't match the signature.

mirabilos:
Antony Messerli dixit:

> Thanks, I gave this a try and it appears to work well.  I'll go ahead
> and add it. It also appears it's referencing the now defunct
> netboot.me if you want to change it.

Yes, the next snapshot will contain netboot.xyz in this place.

> If you guys get releases staged out, we can always add the checksums.
> (just difficult for rolling right now, haven't figured out the easiest
> low-maintenance way for that yet)

No, we switched to rolling because there’s not enough development
nor manpower to do full releases, with QA and all that. But I think
you might wish to look into having multiple signers, then requiring
that images for one group be signed with one specific key, etc. (that
is, always specifying in the boot menu _which_ key is allowed to sign
that particular image, never allowing _all_ keys), then we probably
can just push new signatures.

I can also just send you a message somehow (eMail?) whenever I sync
the new image out (with actual checksums, and PGP signed, or something).

> For the future as an FYI, when testing a remote branch from github,
> all you have to do is disable the signature checks, since it's
> attempting to use the original code signature and the modified file on
> github doesn't match the signature.

OK, thanks!

PR #105, opened 2016-07-15
Adds support for loading Windows Installs

antonym:
Requires the user to specify an HTTP network location of extracted ISO.  This is experimental and it may break.


chris18890:
:+1:  will play around with this over the weekend, very interesting! :)

antonym:
Yeah, let me know how it works.  I may also add some hooks into netboot.xyz-custom so that you can preset your url variables by github user.  I haven't tested yet as I need to upload the bits somewhere to try it out.

PR #109, opened 2016-08-18
Trying to fix path issue with windows installation

netbrain:
This fixes the URL path that failed for me when netboot tried the following:

```
Checking wimboot signature...
http://boot.netboot.xyz/sigs//wimbooy.sig... Operation not permitted
```

Further expection showed that the url gave an error response. Removing the addition / in the URL solved the issue for me.


antonym:
Thanks!

PR #128, opened 2017-01-28
Boot Ubuntu Live System via HTTP

schierlm:
This uses initramfs built by
<https://github.com/live-httpboot/ubuntu-live-httpboot/>, which in turn
uses httpfs2 from <http://httpfs.sourceforge.net> to mount the live ISO
via HTTP.

Kernel and Initamfs are hosted at SourceForge since iPXE does not like
downloads from GitHub.

Due to lack of support for nslookup in current netboot.xyz images, the IP
address for releases.ubuntu.com is hardcoded as a fallback if nslookup is
not available.

Httpboot images are currently only available for 14.04.1, 16.04 and 16.10
- if you need anything else, feel free to open an issue at the GitHub
project mentioned above.

### DIFF
--- a/src/ubuntu.ipxe
+++ b/src/ubuntu.ipxe
@@ -30,11 +30,16 @@ echo Setting mirror to ${ubuntu_mirror}
 clear ubuntu_version
 echo -n Please set enter code name of release: ${} && read ubuntu_version
 set dir ubuntu/dists/${ubuntu_version}/main/installer-${arch_a}/current/images/netboot/
+set httpboot_version NONE
 goto deb_boot_type
 
 :mirrorcfg
 set mirrorcfg mirror/suite=${ubuntu_version} mirror/country=manual mirror/http/hostname=${ubuntu_mirror} mirror/http/directory=/ubuntu
 set dir ubuntu/dists/${ubuntu_version}/main/installer-${arch_a}/current/images/netboot/
+set httpboot_version NONE
+iseq ${ubuntu_version} trusty && set httpboot_version 14.04.1 ||
+iseq ${ubuntu_version} xenial && set httpboot_version 16.04 ||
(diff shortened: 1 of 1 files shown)


antonym:
Is this something that could be upstreamed into the distributions release cycles?  Ideally we could get the distros to extract the kernels and rootfs from each release and hosted in the same locations as the ISOs so that they could be booted using dracut.  Then we could boot the live image much faster.

Loading modified kernel and initrds that are outside of a distributions release cycle is not really ideal and can introduce security risks.

schierlm:
Ideally, yes :-) But we don't live in an ideal world.

To be honest, my experience with https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=724931 (not opened by me but if you scroll down you will see I was also participating in it), which is still open (although it contains more than one patchset that worked for me and is in use by myself for the last few years), made me believe it is easier to fork anything in Debian/Ubuntu based distributions than trying to get it upstreamed (especially as long as the people interested in it are a minority compared to the project user base).

The comment about Live CDs on https://boot.fedoraproject.org/faq (which has been there [unchanged at least since 2010](http://web.archive.org/web/20101121001057/http://boot.fedoraproject.org/faq), although httpfs exists since 2006 and httpfs2 since 2012) makes me believe there is not much interest in netbooting live media upstream (at least not by the developers, more by the users who ask questions)...

If you have better contacts upstream than me, you can of course pursue this way.

I also accept if you don't want to merge it (just one more thing I have to maintain a fork of...), but it was at least worth a try :-)

antonym:
Closing stale PR, using the webboot method in Debian live, would love to see that continue on and become more widely used.

schierlm:
Hmm. Doesn't webboot download the whole image before it can start to boot?

PR #138, opened 2017-04-12
Add ALT Linux Rescue

willglynn:
Thanks for making `netboot.xyz`!

This commit adds [ALT Linux Rescue](https://en.altlinux.org/Rescue), a tool which I've used in the past for various maintenance purposes and used again today to get convenient access to `ipmitool` on a series of diskless machines. This image is larger than I'd like -- about 400 MB -- but it works. I put it in the Utilities menu but there's an argument for the Live menu too. 

One snag: I'm unclear on how image signatures are generated. [`prep-release.sh`](https://github.com/antonym/netboot.xyz/blob/master/script/prep-release.sh) signs the scripts, but something else signs the images… right?

I ask because this changeset points to an image which is regularly rebuilt and for which signatures will therefore regularly become invalid. There are [dated builds available](http://nightly.altlinux.org/sisyphus/snapshots/) which could be used to get consistent binaries, but this comes at the cost of periodically updating the date and regenerating the signatures. I would be happy to script that but I don't know where or how best to make that happen.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -69,6 +69,7 @@ You'll need to make sure to have [DOWNLOAD_PROTO_HTTPS](https://github.com/ipxe/
 
 #### Utilities
 
+* [ALT Linux Rescue](https://en.altlinux.org/Rescue)
 * [AVG Rescue CD](http://www.avg.com/us-en/avg-rescue-cd)
 * [Clonezilla](http://www.clonezilla.org/)
 * [DBAN](http://www.dban.org/)
--- a/src/utils.ipxe
+++ b/src/utils.ipxe
@@ -2,6 +2,7 @@
 
 menu Utilities - Image Sig Checks: [${img_sigs_enabled}]
 item --gap Utilities:
+item alt_linux_rescue ${space} ALT Linux Rescue
 item avg ${space} AVG Rescue CD
 item clonezilla ${space} Clonezilla
 item dban ${space} DBAN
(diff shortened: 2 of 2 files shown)


antonym:
Thanks for the contrib.  I'm using this job to grab the ISOs and generate sigs periodically (https://github.com/antonym/netboot.xyz-sigs/blob/master/generate-utility-sigs.sh) so we'd need to add the url of the release there.  

One issue I see is that they seem to clean up older releases quite often which makes it hard to maintain on this side without changing the release date every few weeks.  It might be good to suggest to the project that they throw some of the more tested releases into the archive directory so that we could pull directly from that or at least keep one of the releases.

We could always mirror the release at https://sourceforge.net/projects/netboot-xyz/?source=directory too but I typically prefer to pull from the project directly so I don't have to maintain builds myself (plus sourceforce is super sluggish).

willglynn:
I understand not wanting to mirror stuff. I'll write the ALT Linux folks and see what they can do.

Another thought: have you considered reaching out to [ibiblio](http://www.ibiblio.org/share/) or [kernel.org](https://www.kernel.org/category/faq.html#i-have-cool-project-x-can-you-guys-mirror-it-for-me) for `netboot.xyz`-related hosting? Either might be willing to mirror things more generally, and presumably both would let you manipulate the hosted files using tools that are more automation-friendly (and CI-friendly!) than SourceForge. On the one hand, it sucks to maintain copies of things -- but on the other, `netboot.xyz` is stuck on every upstream's update treadmill anyway.

willglynn:
I opened ALT Linux [issue #33374](https://bugzilla.altlinux.org/show_bug.cgi?id=33374), explaining the situation and referencing this PR.

antonym:
Looks good, I'll merge this version and we can start testing it.  I've generated sigs here: 

https://github.com/antonym/netboot.xyz-sigs/commit/938728ae36787d57ae2a1000f609807217ec46cd

Ideally we can get away from the ISO and hit the files directly as you mentioned in the issue, it would work much faster and support providers that require serial support as then we can tune the cmdline.

Thanks for your help chasing this down and your contrib!

PR #142, opened 2017-05-03
openbsd: 6.1

ghost:
I were unable to test using the custom branch option.

screen shot 2017-05-03 at 09 51 37

### DIFF
--- a/src/openbsd.ipxe
+++ b/src/openbsd.ipxe
@@ -5,14 +5,14 @@
 
 :openbsd_menu
 menu OpenBSD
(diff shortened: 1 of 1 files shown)


antonym:
Added 6.1:

https://github.com/antonym/netboot.xyz/commit/9625e48922d46ac547f88fa3687e242227b48e79

In the future you can test by disabling netboot.xyz signatures on the main menu.  That should let you proceed to test as the iPXE file you're testing signature would not match the existing signature.

ghost:
Thank you for explaining.  I'll be sure to do that :-)

PR #146, opened 2017-05-18
parrotsec: changed url and version to stable

ghost:


antonym:
Thanks for the update.  Merging in.

I also noticed while verifying this that installer kernels are available which would let me drop the iso and use the better method.  Unfortunately it didn't work out of the box (https://github.com/antonym/netboot.xyz/commit/0e9199a118c053cb4b3c2db97ab857cdf2cf7dba) 
image

Probably something to raise upstream with the project, as using the installer kernels is much quicker.

PR #150, opened 2017-05-21
update some of the util distros & use version variables

chris18890:
update some of the util distros & use version variables


antonym:
Thanks for the contrib!

PR #164, opened 2017-06-30
Update version for live Debian

ghost:
add fallback so that this entry no longer breaks entirely due to having a single dependency on a specific version being available

### DIFF
--- a/src/livedebian.ipxe
+++ b/src/livedebian.ipxe
@@ -9,7 +9,8 @@ goto ${menu}
 set os Debian Live
 menu ${os}
 item --gap Versions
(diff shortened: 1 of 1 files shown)


antonym:
Unfortunately I don't think this will work as the release is no longer on the mirrors.  I made a bug for this:

https://github.com/antonym/netboot.xyz/issues/160

I don't think upstream put out a webboot release.  Need to go someone upstream to get it added to the mirrors, just not sure who that contact might be.

antonym:
Fixed for now to retrieve 8.8.9 from cdimage.

PR #168, opened 2017-08-02
Update documentation on packet.net for persistent PXE boot.

spaletta:


antonym (commented):
Thanks, meant to update the doc with this info, but you beat me to it!

PR #169, opened 2017-08-04
remove unsupported versions

chris18890:


antonym (approved):
LGTM according to https://wiki.ubuntu.com/Releases

If older ones are needed, the codenames can still be specified to boot them.

Thanks!

PR #172, opened 2017-08-24
Punt OpenBSD snapshots to 6.2

job:
OpenBSD is now at 6.2-current: http://undeadly.org/cgi?action=article&sid=20170821133453


job:
OpenBSD is now at 6.2-current: http://undeadly.org/cgi?action=article&sid=20170821133453

antonym:
Thanks!

PR #174, opened 2017-09-05
update to supported FreeBSD releases

lifanov:
This updates FreeBSD installers to 11.1, 11.0, and 10.3, which are versions currently supported upstream. This also changes upstream site to mfsbsd. I tested this with all three installers with x86_64, but unfortunately 10.0-RELEASE was the last release available from mfsbsd with i386 image for it.

### DIFF
--- a/src/freebsd.ipxe
+++ b/src/freebsd.ipxe
@@ -10,10 +10,6 @@ item 11.0 FreeBSD 11.0
 item 10.3 FreeBSD 10.3
 choose ver || goto freebsd_exit
 
-iseq ${ver} 11.1 && set image_ver_major 11 ||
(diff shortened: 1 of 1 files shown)


antonym:
Hey, this looks awesome.  I had been meaning to fix this up for ages but never got around to it.  This should make it much easier to maintain going forward.  Should also be able to look into adding signature checking next on the images as well for the images on the mfsbsd.vx.sk domain.

Thanks again!

lifanov:
No problem! I'm currently investigating what it would take to ship mfsBSD images as part of official FreeBSD release process, which could simplify this a lot and also enable i386 support.

PR #190, opened 2017-10-19
Add Ubuntu Artful Aardvark

carroarmato0:
New release, tested and works just like the other releases in the menu.


antonym:
Woot, thanks!

PR #192, opened 2017-10-29
debian: Use deb.debian.org as mirror

### LINKED ISSUE #137: Change default mirror from rackspace.com to something different
[boot.cfg](https://github.com/antonym/netboot.xyz/blob/master/src/boot.cfg) currently specifies mirror.rackspace.com as the default mirror site. What's annoying about this is that linux distributions provided by Rackspace have modified `/etc/apt/sources.list` files, which again point to mirror.rackspace.com instead of official mirrors.

So how about the default mirror site is set to an official/independent mirror site?

For the time being, I work around this issue by manually changing the  `mirror` variable via the iPXE shell.

nicoonoclaste:
Closes #137

### DIFF
--- a/src/debian.ipxe
+++ b/src/debian.ipxe
@@ -9,7 +9,7 @@ goto ${menu}
 set os Debian
 clear debian_version
 clear older_release
-set debian_mirror ${mirror}
+set debian_mirror deb.debian.org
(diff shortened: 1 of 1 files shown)


antonym:
Will be included via https://github.com/antonym/netboot.xyz/pull/193

Thanks!

PR #197, opened 2017-12-12
Fedora 25 is eol

fabaff:
Fedora 25 is now End-of-life. Thus Fedora 24 is already (August 2017).

As a reference: https://fedoramagazine.org/fedora-25-end-life/


antonym:
Thanks!

PR #208, opened 2017-12-16
Move links out of the text

fabaff:


antonym (changes requested):
Think the links are missing parenthesis so the view is not rendering correctly.

fabaff:
Thanks for the feedback. Should be fixed now.

PR #210, opened 2017-12-18
[WIP] Add IPFire

fabaff:
The Open Source Firewall Distribution
https://www.ipfire.org/

That one is not really following the style of the existing distributions. It feels not very smooth.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ Full documentation is at ReadTheDocs:
 * [FreeBSD](https://freebsd.org)
 * [FreeDOS](http://www.freedos.org)
 * [Gentoo](https://gentoo.org)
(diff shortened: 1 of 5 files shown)


antonym:
Good find, it looks like we can add it.  We can probably skip all the loading and load the bits directly using the sources from here:

http://boot.ipfire.org/menu.cfg?arch=x86_64&platform=pcbios

antonym:
Yeah, your latest update is along the lines of what I was thinking of doing, just pointing at the kernels directly and bypassing all of their menus.  Let me know once it's ready to review and I'll kick the tires on it.

Thanks!

antonym:
Closing since I put together an updated PR.  Thanks for your work on this!

https://github.com/antonym/netboot.xyz/pull/221

PR #212, opened 2017-12-20
Adds support for Fedora Atomic

antonym:
Adds basic support for using the Fedora 26/27 Atomic pxe boot images.


fabaff:
Works for me in QEMU/KVM.

For maintenance reasons I would like to suggest that we replace the date (incl. the minor release) from `dir` with a var. Could be done with the first update, though.

antonym:
I'll probably just leave it as the URL for now because adding the specific version would probably confuse things.  Hopefully, this format is temporary too and we'll see the date code go away eventually.

Issue #30, opened 2016-01-13
AlpineLinux Support


zQueal:
For anyone who doesn't know, Alpine Linux is tiny. The standard vanilla iso is 81MB with the extended iso a measly 360MB installed.

It would make a really great addition to netboot. They offer a public download repository [here](http://dl-3.alpinelinux.org/alpine/).

antonym:
I gave this a try tonight but unfortunately Alpine Linux ISOs do not play nicely with memdisk which is what I need in order to boot.  You can read more about the issue [here](https://www.reversengineered.com/2016/01/07/booting-linux-isos-with-memdisk-and-ipxe/) and a potential way to fix it.  It might be worth opening the ticket with them and seeing in they are willing to implement the above item.

The way I usually validate an ISO works or not is pulling up the iPXE command shell and running:

kernel http://boot.netboot.xyz/memdisk iso raw
initrd http://URL.to/ISO
boot

If it works, it'll usually start the install or load up all of the way.  If it fails, it'll usually fail early after being unable to load the rootfs image.

The ISOs can be cracked open and hosted somewhere to boot, but that introduces a lot of other issues. (tampering, bandwidth, space, etc).  I'd much rather pull from the provider of the operating system or a trusted mirror if at all possible and leave netboot.xyz as a router to those operating systems.

antonym:
Looks like some others have tried as well, might be a good place to start. http://lists.alpinelinux.org/alpine-devel/3292.html

zQueal:
I've been trying to get some involvement via the Alpine official IRC, but there's been no movement in more than 4 hours despite over 300 connected clients. Not entirely sure this is going to happen. It's far outside my skillset.

andyshinn:
Alpine Linux also has a bug tracker at https://bugs.alpinelinux.org/ and a couple mailing lists I would try as well: http://lists.alpinelinux.org/.

zQueal:
Andy, thanks!

I've opened a thread on their [internal forums](http://forum.alpinelinux.org/forum/installation/netbootxyz-compatability).

antonym:
Thanks, guess we just wait and see.  If we can get a working image, it'll be really easy to get it going.

bfritz:
[...]

PR #223, opened 2018-02-20
add Alpine Linux

clandmeter:
This is a first attempt of adding Alpine Linux to netboot.xyz.

We currently do not "officially" ship netboot images but this is high on my prio list. Instead I've added images for both x86 and x86_64 from both stable and edge releases to our dev server. I also have a few small patches ready to go into Alpine Linux to fix some corner cases which will be applied asap.

This has been tested with my custom [ipxe script ](https://github.com/clandmeter/netboot.xyz-custom/blob/master/custom.ipxe). Please check if i missed anything.

ref: #30

### DIFF
--- a/src/alpinelinux.ipxe
+++ b/src/alpinelinux.ipxe
@@ -1,0 +1,29 @@
+#!ipxe
+
+# Alpine Linux
+# https://alpinelinux.org
+
+goto ${menu}
+
+:alpinelinux
+set os Alpine Linux
+iseq ${arch} x86_64 && set bootarch x86_64 || set bootarch x86
+
+menu ${os} [${bootarch}]
+item latest-stable Latest stable
+item edge Edge (development)
+choose version || goto alpine_exit
+goto boot
+
+:boot
+set base-url http://dev.alpinelinux.org/~clandmeter/netboot/${version}/${bootarch}
+set repo-url http://${alpinelinux_mirror}/${alpinelinux_base_dir}/${version}/main
+imgfree
+kernel ${base-url}/vmlinuz-vanilla alpine_repo=${repo-url} modules=loop,squashfs modloop=${base-url}/modloop-vanilla quiet nomodeset
+initrd ${base-url}/initramfs-vanilla
+boot
+goto alpine_exit
+
+:alpine_exit
+clear menu
+exit 0
--- a/src/linux.ipxe
+++ b/src/linux.ipxe
@@ -15,6 +15,7 @@ item manjaro ${space} Manjaro Linux
 item opensuse ${space} openSUSE
 item ubuntu ${space} Ubuntu
 item --gap All Others:
+item alpinelinux ${space} Alpine Linux (Experimental)
 item antergos ${space} Antergos
 item coreos ${space} CoreOS
 item devuan ${space} Devuan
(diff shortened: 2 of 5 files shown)


src/alpinelinux.ipxe
```diff
@@ -0,0 +1,29 @@
+#!ipxe
+
+# Alpine Linux netboot.xyz ipxe script.
```
antonym: Can just leave this as Alpine Linux to make it consistent with the other iPXE files.

src/linux.ipxe
```diff
@@ -6,6 +6,7 @@ goto ${menu} ||
 menu Linux Installers - Current Arch [ ${arch} ]
 iseq ${arch} x86_64 && set arch_a amd64 || set arch_a ${arch}
 item --gap Popular Linux Operating Systems:
+item alpinelinux ${space} Alpine Linux
```
antonym: Can we move it down to the All Others section and mark it as *Experimental*?

src/alpinelinux.ipxe
```diff
@@ -0,0 +1,29 @@
...
+iseq ${arch} x86_64 && set bootarch x86_64 || set bootarch x86
+
+menu ${os} [${bootarch}]
+item latest-stable Latest stable
+item edge Edge (development)
+choose version || goto alpine_exit
+goto boot
+
+:boot
+set base-url http://dev.alpinelinux.org/~clandmeter/netboot/${version}/${bootarch}
```
antonym, line 19: I'm hesitant to add unofficial images but I'm ok with treating the distro as experimental as long as they are hosted from the alpinelinux.org domain.  If you can make progress on official images, we can remove the experimental tag and add signature checking for the images.  That way we can at least let the users experiment with the distro.

antonym (changes requested):
This looks great so far, I tested it in VBox and it worked well.  I have a few comments and changes requested.

clandmeter:
All sounds good to me. I made the changes you requested.

Regarding the signatures, how frequent do you verify if the images are updated?
Is it possible users could have broken sigs if we update the images before you re-sign them?

antonym:
@clandmeter Thanks, looks good.  On the signatures, I currently try to avoid the releases that reside in a "rolling" directory and try and pull images or kernels from a location that is usually static and tagged as a release.  This way I don't have to worry about things breaking if there's a version bump which saves me having to chase things down quite a bit.  Currently the signature generation isn't the greatest and needs some rework but the script I use is here:

https://github.com/antonym/netboot.xyz-sigs/blob/master/generate-linux-sigs.sh

It goes and retrieves all of the latest kernels, signs them, and the keeps the signature files.  I plan on reworking the system into something a little less ghetto at some point, but it's a little fun to manage given the number of versions out there.

antonym (approved):

antonym:
Merging this in for now, but keep us updated on getting official builds.  Thanks!

clandmeter:
@antonym we now support netboot from repos since our latest release v3.8.

You can take a look at http://dl-cdn.alpinelinux.org/alpine/v3.8/releases/

antonym:
Yep, saw that, will try and push up an update this week for it.  Thanks!

PR #225, opened 2018-03-17
Update utils.ipxe - Breakin

Firminator:
* Added note that it supports only amd64/x64. Booting on a i686 machine will fail as this tool was build for servers and clusters in mind, not for old desktops/laptops.
* Added version and description: v4.26.1-53 (2015-10-22) - Hardware stress-test 
* Added comments for website and download link.
They have a PXE image as well. Can iPXE download http://www.advancedclustering.com/wp-content/uploads/2017/02/bootimage-4.26.1-53.tbz2 and untar on the fly and then load kernel and initrd?


antonym:
Thanks for the contrib, closing this though as it adds a bit of unnecessary info to the menu option.

PR #226, opened 2018-03-17
Update ipfire.ipxe

Firminator:
* Added latest release
* Added dates
* Added Website and Download links

How do you change the boot architecture from x86_64 to i586 so that the download path can be adjusted? I see different approaches as I try to understand the code... plus we have i386, i686, and in this case i586. Would the archlinux.ipxe be a working and up to date template for the boot architecture variable determination (iseq ${arch} i386 && set bootarch i586 || set bootarch x86_64)?

I think `|| goto linux_menu` does cause an error and needs to be removed. Not sure though.

### DIFF
--- a/src/ipfire.ipxe
+++ b/src/ipfire.ipxe
@@ -1,15 +1,22 @@
 #!ipxe
 
-# IPFire
-# https://www.ipfire.org/
+# IPFire - Open Source Firewall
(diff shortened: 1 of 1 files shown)


antonym:
Closing PR out as it's out of date, probably don't need to manage multiple versions and we should be able to stick to just the latest.

PR #233, opened 2018-03-20
Add latest versions of Arch

fabaff:


Firminator:
Not sure when they introduced it, but they actually have a 'latest' branch which is great, as this removes the need to manually update the archlinux.ipxe file each month:
https://mirror.wormhole.eu/archlinux/iso/latest/arch/boot/x86_64/
or
even on rackspace: http://mirror.rackspace.com/archlinux/iso/latest/arch/boot/x86_64/

fabaff:
Doesn't really help, from my point of view, as it doesn't work for the older releases. But yes, if we only allow the latest releases then `latest` is a great way to do it. Personally, I like to have an option to choose but this could only be me.

Firminator:
Yes someone pointed out in one of the other threads I was reading last week that they want an easy way to go back in releases. So fair enough.
However Arch is Rolling Release... as soon as you update it for the first time it should be 'latest', so keeping the older release might really be a moot point. Keeping older release might be more feasible for non-rolling-releases distros. Just pointing this out. Both ways is fine with me.

antonym:
Yeah, the only bad thing about using latest is having to validate image matches the signature as the image changes out from under you monthly and causes a signature failure.  It's definitely the easier path though but I'd have to disable signature checking for Arch then.

Firminator:
I might be missing the big picture here, or a detail, but arch provides sigs in the same folder:
http://mirror.rackspace.com/archlinux/iso/latest/arch/boot/x86_64/archiso.img.ipxe.sig
and
http://mirror.rackspace.com/archlinux/iso/latest/arch/boot/x86_64/vmlinuz.ipxe.sig

We can't use these?

antonym:
Arch generates their own iPXE binaries (https://www.archlinux.org/releng/netboot/) that include the cert that they sign those signatures with.  netboot.xyz uses its own cert to sign for many other operating systems so I don't think we could use theirs to validate.  I could chain into their kernel but it changes the experience up especially if the user decides they want to back out of the Arch menu.

antonym (approved):

PR #241, opened 2018-05-22
Updates

TimmEDK:
Some new versions and path

### DIFF
--- a/src/manjaro.ipxe
+++ b/src/manjaro.ipxe
@@ -25,7 +25,7 @@ goto mirror_cfg
 :mirror_cfg
 set flavor_root ${flavor}
 iseq ${flavor} architect && set ${manjaro_version} 17.1.10 ||
-set manjaro_initrd http://ftp.acc.umu.se//mirror/osdn.net/storage/g/m/ma/manjaro/${flavor_root}/manjaro-${flavor}-${manjaro_version}-stable-x86_64.iso/download
+set manjaro_initrd http://ftp.acc.umu.se//mirror/osdn.net/storage/g/m/ma/manjaro/${flavor_root}/manjaro-${flavor}-${manjaro_version}-stable-x86_64.iso
 goto manjaro_boot
 
 :manjaro_boot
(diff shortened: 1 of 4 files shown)


src/manjaro.ipxe
```diff
@@ -25,8 +24,8 @@ goto mirror_cfg
 
 :mirror_cfg
 set flavor_root ${flavor}
-iseq ${flavor} architect && set ${manjaro_version} 17.1.7 ||
-set manjaro_initrd http://sourceforge.net/projects/manjarolinux/files/release/${manjaro_version}/${flavor_root}/manjaro-${flavor}-${manjaro_version}-stable-x86_64.iso/download
+iseq ${flavor} architect && set ${manjaro_version} 17.1.10 ||
```
antonym, line 27: Looks like Architect is 17.1.9 still.

src/manjaro.ipxe
```diff
@@ -9,13 +9,12 @@ goto ${menu}
...
 menu ${os} Installers
 item --gap Official Releases
-item 17.1.8 ${space} ${os} 17.1.8
+item 17.1.10 ${space} ${os} 17.1.10
 choose manjaro_version || goto manjaro_exit
 
 :manjaro_flavor
 menu ${os} ${manjaro_version} Installers
 item --gap Flavors
-item architect ${manjaro_version} Architect
```
antonym, line 18: Any reason for dropping?

antonym (changes requested):
Great updates, any reason for dropping Architect?

src/manjaro.ipxe
TimmEDK: Only because there aren't no 17.1.10 for that official

antonym: Yeah, Architect isn't always updated every version, that's why the iseq for arcitect is in place, if you  can change it to:

```
iseq ${flavor} architect && set ${manjaro_version} 17.1.9 ||
```
And add back in Architect we can see about merging it.

Thanks!

TimmEDK: Sorceforge has switched to osdn, which doesn't fetch the ISO in the script - even directly

antonym: Yeah, I see the issue, we'll probably need to dig into that as it's currently using Sweden mirrors which aren't ideal.  It's broken currently anyways so we can move it in and we can open a bug for optimizing that mirror.

antonym (approved):

PR #244, opened 2018-06-27
Update antergos.ipxe

bitmux:
Antergos's own link to "latest" minimal iso download. I don't know that this will actually work..


antonym:
Yeah, the redirect is a little flaky so I'd probably avoid using it.  Pointing directly to the ISO URL is usually a bit more safer and reliable.

PR #246, opened 2018-06-30
Update Alpine Linux to new netboot images

antonym:
Update Alpine Linux to new netboot images and add signature checks to Alpine.


nekocentral:
Just tested it, booted straight up and fast but noticed during install it skipped the part to select the disk to install but that might be a alpine related thing.
Thanks for adding it

PR #254, opened 2018-08-02
Fix modloop URL in alpinelinux.ipxe

galexite:
The URL was missing the `/${dir}`, resulting in a 404 when Alpine tries to fetch the modloop, thus preventing extra kernel modules from being loaded.


antonym:
Thanks!

PR #261, opened 2018-08-23
Upgrade Manjaro to 17.1.12

fabaff:


src/manjaro.ipxe
```diff
@@ -24,7 +24,7 @@ goto mirror_cfg
 
 :mirror_cfg
 set flavor_root ${flavor}
-iseq ${flavor} architect && set ${manjaro_version} 17.1.10 ||
+iseq ${flavor} architect && set ${manjaro_version} 17.1.12 ||
```
antonym: Think Architect is still on 17.1.11, don't think they have a 17.1.12 version yet.

antonym (approved):

src/manjaro.ipxe
fabaff: Thanks for the feedback. You are right, it's still 17.1.11 for Architect.

PR #264, opened 2018-08-23
Upgrade Debian Live to 8.11.0

fabaff:


antonym:
Don't think they've come out with the weboot files for 9.5.0.  In fact I think the current version doesn't have the webboot files there anymore either, so the Live might actually be broken.

fabaff:
8.11.0 seems to be last release were they provided the webroot files.

antonym:
Cool, thanks!

antonym (approved):

PR #265, opened 2018-08-23
Upgrade IPFire to 2.21 Core 122

fabaff:
Should perhaps be merged with #226.

### DIFF
--- a/src/ipfire.ipxe
+++ b/src/ipfire.ipxe
@@ -9,7 +9,7 @@ goto ${menu} ||
 clear osversion
 set os IPFire
 menu ${os} - Image Sig Checks: [${img_sigs_enabled}]
-item 2.19-core118 ${space} ${os} 2.19 Core 118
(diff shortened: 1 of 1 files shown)


src/ipfire.ipxe
```diff
@@ -9,7 +9,7 @@ goto ${menu} ||
 clear osversion
 set os IPFire
 menu ${os} - Image Sig Checks: [${img_sigs_enabled}]
-item 2.19-core118 ${space} ${os} 2.19 Core 118
+item 2.19-core118 ${space} ${os} 2.21 Core 122
```
antonym: Make sure to change the second field too. (item)

antonym (changes requested):

src/ipfire.ipxe
fabaff: Thanks for the feedback.

antonym (approved):

PR #272, opened 2018-10-19
Update OpenBSD menu options

doublej472:
A new version of [OpenBSD](https://www.openbsd.org/64.html) was released yesterday! When trying to test it out with netboot.xyz, I noticed that the menu options needed to be updated, which the PR addresses. This PR also updates the snapshot entry since it was outdated.


antonym:
Thanks for the contrib!

antonym (approved):

PR #275, opened 2018-10-20
remove XenServer 6.5 & update 7.0 > 7.6

chris18890:


antonym:
Probably going to end up dropping XenServer as I don't have a good place to host the extracted bits as Citrix was putting them on their CDN previously.  They made a change in some of the later versions which broke accessing XenServer from CDN (was looking for a specific .treeinfo file in the directory).

antonym:
Closing since XenServer as an option was dropped.

PR #277, opened 2018-10-24
AlpineLinux: Update version number

doublej472:
[Alpine Linux](https://www.alpinelinux.org/posts/Alpine-3.8.1-released.html) has been updated to 3.8.1! This PR doesn't actually change how Alpine is booted (netboot.xyz already boots the latest 3.8 version automatically), but it updates the version number that the user sees to avoid confusion.


antonym:
Thanks!

antonym (approved):

PR #278, opened 2018-10-24
Update Centos 6.9 -> 6.10

doublej472:
This updates the minor version of Centos 6.9 to 6.10. I'm not sure if anyone is still installing Centos 6 for new servers, but it wont be EOL until [November 30, 2020](https://wiki.centos.org/FAQ/General#head-fe8a0be91ee3e7dea812e8694491e1dde5b75e6d), so we might as well keep a boot option for it.

Tested on my T61 Thinkpad.


antonym (approved):

antonym:
Thanks!

PR #303, opened 2019-02-06
Update archlinux to 2019.02.01

doublej472:
There's a new image for Archlinux, which this PR updates

I've tested this on my laptop and it boots fine


antonym (approved):
Thanks!

PR #304, opened 2019-02-06
Adds NixOS

Lassulus:
Add nixos
related: https://github.com/antonym/netboot.xyz/issues/37

### DIFF
--- a/src/linux.ipxe
+++ b/src/linux.ipxe
@@ -21,6 +21,7 @@ item coreos ${space} CoreOS
 item devuan ${space} Devuan
 item gentoo ${space} Gentoo Linux
 item ipfire ${space} IPFire
+item nixos ${space} NixOS
 item rancheros ${space} RancherOS
 item rhel ${space} Red Hat Enterprise Linux
 item scientific ${space} Scientific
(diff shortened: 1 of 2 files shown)


antonym:
Looks good, it loaded right up for me in Virtual Box.  Thanks for making the PR.  It appears to be pulling the completed artifacts from a build server.  Is this the desired permanent location or are there plans to move these to a release mirror that might be faster?  If not, that's fine, we can always update the URL, just want to make sure we're pulling from an acceptable location.  We can also work on getting the binaries signed next as well so that we can add a level of trust to them too.

Thanks again!

Lassulus:
well there was an idea to add them to the release-channel https://github.com/NixOS/nixos-channel-scripts/pull/21 but this got delayed for unknown reasons. This location should be permanent, we will have to change the stable channel every 6 months, but that should be straightforward.

Would be happy to work on the signing if you could poke me in the right direction :)
thanks!

antonym:
Sounds good, I'll go ahead and merge and we can close this out.  The signing stuff is a bit dirty and I'm probably going to rework it a bit so we can include that at that time.

antonym (approved):

vcunat:
You didn't modify readme :-)  (Just noticed, certainly not a big deal for me.)

PR #310, opened 2019-02-23
Add NixOS to README.md

Lassulus:
as pointed out by vcunat in https://github.com/antonym/netboot.xyz/pull/304#issuecomment-466630702


antonym (approved):
Thanks!

PR #314, opened 2019-03-18
Upgrade Arch Linux to 2019.03.01

fabaff:


Centzilius (approved):

Centzilius:
On a different note why aren't we using the `latest` folder so this is always up to date?

antonym:
I'd have to drop the signature checking on Arch if I pointed to latest as the signature would break monthly which is why I point to the version explicitly.

antonym (approved):

PR #319, opened 2019-03-29
Update ubuntu.ipxe

chris18890:
removes 14.04 & adds 19.04

Merge after 2019-04-18 as per https://en.wikipedia.org/wiki/Ubuntu_version_history#Table_of_versions when disco is released


antonym (approved):
Thanks!  Trusty goes EOL this month so LGTM.

PR #321, opened 2019-04-15
Fix Fedora-Atomic-29 URL

### LINKED ISSUE #320: Fedora Atomic URL broken
# Problem statement
It seems like the current URL for atomic is broken:
Auswahl_146

# Solution proposal
Directly - this would mean we need to update:
https://github.com/antonym/netboot.xyz/blob/25f936aee5c800f15b3ca10f057444026cc39767/src/fedora.ipxe#L29

The latest URL is:
`http://mirrors.kernel.org/fedora-alt/atomic/stable/Fedora-29-updates-20190401.0/AtomicHost/x86_64/os/images/pxeboot/`

## Additional enhancements
Also we should update the fedora mirror:
https://github.com/antonym/netboot.xyz/blob/924f7d746459ee879b08ec0c73027ec59e967e16/src/boot.cfg#L50
Using `https://download.fedoraproject.org/pub/alt/` is the same as `http://mirrors.kernel.org/fedora-alt/` - but it will randomly chose a close-by mirror.

mulbc:
Fix URL so that we can continue installing Fedora Atomic 29

Fixes https://github.com/antonym/netboot.xyz/issues/320

### DIFF
--- a/src/fedora.ipxe
+++ b/src/fedora.ipxe
@@ -26,7 +26,7 @@ item Workstation ${ova} Workstation
 item Atomic ${ova} Atomic
 isset ${sku_type} || choose sku_type || goto fedora
 set dir ${fedora_base_dir}/releases/${osversion}/${sku_type}/${arch}/os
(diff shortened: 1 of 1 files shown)


antonym (approved):
Thanks!

PR #329, opened 2019-04-30
Upgrade to Fedora 30

fabaff:
https://fedoramagazine.org/announcing-fedora-30/


antonym (approved):
Already updated for F30, but we can remove 28.

PR #336, opened 2019-06-23
Add Flatcar Linux as a boot option

andrewl3wis:
Adds the fork of CoreOS called [Flatcar Linux](https://www.flatcar-linux.org/) as a boot option. It also adds a release target that is unique to flatcar called "Edge" as a deploy target. 

I debated adding this into the Menu as a set of sub options for CoreOS since they are so closely related, but ultimately decided that this made more sense to split it out, particularly in light of the uncertain future of CoreOS as it transition to Fedora CoreOS, and what is a completely different OS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ Full documentation is at netboot.xyz:
 * [Debian](https://debian.org)
 * [Devuan](https://devuan.org)
 * [Fedora](https://fedoraproject.org)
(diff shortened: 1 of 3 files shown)


antonym:
Nice, will try and find some time to review/test this week, thanks!

antonym (approved):
LGTM

Issue #268, opened 2018-09-27
EFI image doesn't work over libvirt + kvm/qemu


renich:
I'm using Fedora 28 as a host. I tried using UEFI (TianoCore) and it doens't work with your image; it drops to a shell (efilinux).

On the other hand, I try the Fedora 29 Beta ISO image and it works fine. 

Might there be something missing in the EFI iso image?

patrakov:
The EFI iso image is just wrong. It does not contain any El Torito records. Instead, it contains BOOT/EFI/BOOTX64.EFI file, which UEFI does not even look at on optical media.

For now, you can create your own EFI iso by running these commands (assuming that you have dosfstools, cdrecord and mtools installed):

```
wget https://boot.netboot.xyz/ipxe/netboot.xyz.efi
mkdir cdrom
dd if=/dev/null of=cdrom/image.flp seek=2880
mkfs.vfat -M 0xf8  -f 1 cdrom/image.flp
mmd -i cdrom/image.flp ::efi ::efi/boot
mcopy -i cdrom/image.flp netboot.xyz.efi ::efi/boot/bootx64.efi
mkisofs -o ipxe-uefi.iso  -eltorito-platform 0xef -eltorito-boot image.flp -no-emul-boot cdrom
```

I will create a pull request later.

antonym:
Closing this as @patrakov was kind enough to fix.  Please reopen if it's still an issue.

PR #341, opened 2019-07-07
Generate a working EFI ISO image (#268)

patrakov:
Tested in KVM with OVMF.


antonym:
Cool, will check it out this week when I get some free time, thanks!

patrakov:
For the record, I have tested the result not only with KVM, but also on a real server hosted by [Bacloud](https://www.bacloud.com/), over IPMI virtual media redirection.

antonym:
Merged these fixes in with another commit ( https://github.com/antonym/netboot.xyz/pull/345 ) as there were some tweaks that needed to be made to your instructions on Xenial.  Making sure I don't break too much stuff.  Thanks for your contrib and please test the disks and let me know if they work for you.  I tested them in Virtual Box with EFI boot and they seemed to boot fine.

patrakov:
I have downloaded the EFI ISO from netboot.xyz site and verified that it works in KVM.

Issue #342, opened 2019-07-10
CentOS 7 fails to install if a static IP is required


patrakov:
In the CentOS 7 ipxe file, "ip=dhcp" is hard-coded as a kernel option. This is a limitation, because, using a failsafe menu (m), you can in fact set a static IP address. Therefore, ideally, the ip=... parameter for the installer kernel should be generated on the fly from the information that the user entered when statically configured the network.

xmovu:
Facing the same issue, had reported this before @patrakov .

No dice, it just won't work, did you find any solution? Perhaps @antonym can have a second look.

xmovu:
Fixed with #343 commits.

PR #343, opened 2019-07-13
Static IP support

patrakov:
See #342 but this pull request is not limited to CentOS.

I have verified that all Linux distributions offered by linux.ipxe. Status with the static IP configured via the failsafe menu:
 - Alpine boots correctly only if the first network card is used
 - Arch relies on http://ip.address.of.mirror/ to work (bug reported upstream)
 - Gentoo and IPFire do not boot at all, also before my patches
 - RHEL not tested
 - Everything else in the menu works

Regarding Gentoo, the snapshot offered is very old, and it is possible to [use third-party live/rescue Linux systems to install Gentoo](https://wiki.gentoo.org/wiki/Installation_alternatives#Installation_from_non-Gentoo_LiveCDs) anyway. So - maybe drop it?

Regarding rescue tools offered by netboot.xyz, I have not tested them. It's possible that they require similar fixes. However, before working on them, I would like to confirm whether the general approach is OK.

Oh, and there seems to be a variable, ${netcfg}, in Debian-related iPXE files, but I can't find a place where it is set.

### DIFF
--- a/src/opensuse.ipxe
+++ b/src/opensuse.ipxe
@@ -2,6 +2,60 @@
 
 # OpenSUSE Operating System
 # http://opensuse.org
+
+isset ${dhcp-server} || goto static_ip
+set netsetup netsetup=dhcp
+goto goto_menu
+
+:static_ip
+# Need to convert netmask into prefix, because otherwise the installer
+# accepts it but configures the network with /32 prefix, which installs
+# fine but breaks connectivity to devices in the same network.
+set prefix 32
+
+iseq ${netmask} 0.0.0.0 && set prefix 0 ||
+
+iseq ${netmask} 128.0.0.0 && set prefix 1 ||
+iseq ${netmask} 192.0.0.0 && set prefix 2 ||
+iseq ${netmask} 224.0.0.0 && set prefix 3 ||
+iseq ${netmask} 240.0.0.0 && set prefix 4 ||
+iseq ${netmask} 248.0.0.0 && set prefix 5 ||
+iseq ${netmask} 252.0.0.0 && set prefix 6 ||
+iseq ${netmask} 254.0.0.0 && set prefix 7 ||
+iseq ${netmask} 255.0.0.0 && set prefix 8 ||
+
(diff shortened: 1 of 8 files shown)


NiKiZe:
Gentoo is on my todolist to fix, since it now have proper support to be PXE booted, (since https://gitweb.gentoo.org/proj/genkernel.git/commit/?id=24ad5065fa856389ee9b058f57adffbe752da157)

But I should probably also nag them about hosting files separately

xmovu (approved):

xmovu:
Just tested it, definitely works.

antonym:
@patrakov Reviewed it all and it looks good, I haven't had time to test it yet but the approach looks correct and we have a +1.  Thanks for tackling this!

I've been trying to figure out a good method for Gentoo as I'd like it on the menu.  I'd prefer that Gentoo would just host those bits somewhere officially so maybe @NiKiZe knows the right people to poke for that.  But I'll probably get it removed for now until we can actually get it working.  I could build a job that could periodically extract the ISO but I don't really want to get into the business of managing OS installation bits as they should come directly from the distribution.

In general I'd like to probably start moving away from distros or utilities that use ISOs as they just aren't very reliable and work half the time.  I'd love to see distributions just offer installer kernels for their OS or host the latest extracted ISO but that's probably a ways off.

The ${netcfg} stuff might be from some older iterations and those variables probably just need to be removed.  I didn't see it really being called or set anywhere.

antonym (approved):

src/opensuse.ipxe
```diff
@@ -2,6 +2,60 @@
 
 # OpenSUSE Operating System
 # http://opensuse.org
+
+isset ${dhcp-server} || goto static_ip
+set netsetup netsetup=dhcp
+goto goto_menu
+
+:static_ip
+# Need to convert netmask into prefix, because otherwise the installer
```
antonym, line 11: This could probably be moved into a common library or might even be useful in iPXE itself (if it isn't already there) but it's fine for now.

PR #349, opened 2019-08-09
Arch Linux updates

lilyanatia:
* add 2019.08.01 and 2019.07.01 releases
* use HTTPS for downloads

### DIFF
--- a/src/archlinux.ipxe
+++ b/src/archlinux.ipxe
@@ -24,6 +24,8 @@ clear arch_version
 menu ${os} - ${arch} - Image Sig Checks: [${img_sigs_enabled}]
 menu Arch Linux
 item --gap Latest Releases
+item 2019.08.01 2019.08.01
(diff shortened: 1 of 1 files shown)


antonym:
We can probably just leave the latest one drop the older ones.  Also, I'd drop the https as not all mirrors provide https support.

lilyanatia:
the configuration in boot.cfg doesn't use all mirrors. it just uses mirror.rackspace.com, which does support HTTPS.

GlitchedAxiom:
Version 2019.06.01 which is the only option currently, is not available anymore at rackspace mirrors. Could you please add the newest 2019.09.01 version and merge this?

Thanks.

antonym:
Pushed here: https://github.com/antonym/netboot.xyz/commit/5269f76b830378c9af9c71d9d90da3b154a29f8c

Closing out

PR #354, opened 2019-08-29
Add Fedora rawhide

TheMuso:
Only available if image signature checks are disabled, and for x86_64.

Rawhide is updated every day, and is a perpetual release, similar to Debian unstable. As such, image validation is not feesable. I would still like to have rawhide available as an option to easily spin up a VM running it, the only question is the UI.

This change currently shows rawhide when image signature validation is disabled. Would you prefer if rawhide is always shown, and add a special case to disable image validation if rawhide is selected?

### DIFF
--- a/src/fedora.ipxe
+++ b/src/fedora.ipxe
@@ -17,6 +17,7 @@ menu Fedora - ${arch} - Image Sig Checks: [${img_sigs_enabled}]
 item --gap Latest Releases
 item 30 ${space} ${os} 30
 item 29 ${space} ${os} 29
+iseq ${arch} x86_64 && item rawhide ${space} ${os} rawhide ||
 isset ${osversion} || choose osversion || goto linux_menu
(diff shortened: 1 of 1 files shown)


antonym:
You can probably just add it to be visible to all and then have it automatically skip the sigs:

```
iseq ${osversion} rawhide && goto skip_sigs ||
```
right before:
``` 
iseq ${img_sigs_enabled} true && goto verify_sigs || goto skip_sigs
```

I'd probably put the option at the end instead of being first too.  I'm ok with how you have it too, but I'm thinking it would be better to just have it visible, otherwise it's not very obvious how to make it appear without going through the code.

TheMuso:
Ok, your suggested changes have been made, and I agree with you re visibility.

antonym (approved):
Looks good, thanks for the contribs!

TheMuso:
You're welcome.

PR #357, opened 2019-09-28
Adds Mageia 7

maortegar:


src/mageia.ipxe
```diff
@@ -11,6 +11,7 @@ goto ${menu} ||
 :mageia
 set os Mageia
 menu Mageia - ${arch} - Image Sig Checks: [${img_sigs_enabled}]
+item 7 ${os} 7
 item 6 ${os} 6
 item 5 ${os} 5
 item 4 ${os} 4
```
fabaff: We should remove 4 and 5 because they are no longer supported.

fabaff (changes requested):

src/mageia.ipxe
forkiesassds: 6 is also not supported. The official website says this `Mageia 6 will be supported until September 30th, 2019.` Source: https://www.mageia.org/en/support/

antonym: Also, is it now on 7.1?

http://mirror.dacentec.com/mageia/distrib/7.1/x86_64/install/images/Mageia-7.1-netinstall-x86_64.iso

maortegar: I have added 7.1 and removed all other releases.

antonym (approved):

PR #361, opened 2019-10-07
fix nameserver field for centos.ipxe

truatpasteurdotfr:
Hi,

the static DNS is defined by nameserver= not in the ip= field
I hope that syntax is ok!

Same change would be required for scientific.ipxe

Cheers

Tru


chris18890 (approved):

antonym:
Thanks, I pushed this up in another commit along with a few other occurances I found.

PR #364, opened 2019-10-08
Splits Linux and Utility menus for arm64

antonym:
Detects if arm architecure and loads arch specific menu
for linux and utility that reflect options that have potential to work
with arm.

### DIFF
--- a/src/boot.cfg
+++ b/src/boot.cfg
@@ -85,11 +85,14 @@ goto architectures_end
 :x86_64
 goto architectures_end
 :arm64
+set menu_linux 0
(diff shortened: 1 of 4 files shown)


forkiesassds (approved):
Can't test it due to the fact there are no IPXE images for aarch64 that I can run on QEMU, from the looks of it might be a good change for aarch64 users.

antonym:
Yeah, this mainly sets things up to be tested from an ARM machine or from a packet arm instance without too much hassle. Each OS still needs to be worked through to find the proper loading procedure.

PR #371, opened 2019-10-19
add missing iseq line for eoan

forkiesassds:


antonym:
Thanks, reason for this is that when the latest version comes out they don’t have the -updates kernel immediately.  We can probably drop the disco line if disco-updates exists now too.

src/ubuntu.ipxe
```diff
@@ -10,7 +10,7 @@ set os Ubuntu
 clear ubuntu_version
 menu ${os} - ${arch_a} - Image Sig Checks: [${img_sigs_enabled}]
 item --gap Latest Releases
-item eoan ${space} ${os} 19.10 Eoan Ermine
+iseq ${arch} x86_64 && item eoan ${space} ${os} 19.10 Eoan Ermine ||
```
antonym: Shouldn’t need this line

src/ubuntu.ipxe
```diff
@@ -33,6 +33,8 @@ goto deb_boot_type
 set mirrorcfg mirror/suite=${ubuntu_version}
 set dir ${ubuntu_base_dir}/dists/${ubuntu_version}-updates/main/installer-${arch_a}/current/images/netboot/
 iseq ${ubuntu_version} disco && set dir ${ubuntu_base_dir}/dists/${ubuntu_version}/main/installer-${arch_a}/current/images/netboot/ ||
```
antonym: Can drop the iseq line for disco

antonym (changes requested):

src/ubuntu.ipxe
forkiesassds: 19.10 does not come with a installer-i386 folder. As i386 support was dropped.

src/ubuntu.ipxe
forkiesassds: dropped the line for disco

src/ubuntu.ipxe
antonym: Ah ok, fair enough, lgtm.

antonym (approved):

PR #375, opened 2019-10-26
Start of Defining a localized version TFTP boot parameter

thelamer:
Hello from the Linuxserver.io team. First off wanted to let you know we are pushing a simple image with your pxe/efi bins in it with a TFTP server: 
https://hub.docker.com/r/linuxserver/netbootxyz
We also plan on putting a blog entry up as internally team members have found it very useful inside of their VM environments. 

I would like to get the ball rolling on building logic that allows users to self host not only menu files but also kernel/initrd/iso files and serve them via tftp/http. I am only assuming you are interested in this based on https://github.com/antonym/netboot.xyz/issues/26 . 

For right now this PR is only to allow local ipxe files to be served from a TFTP server, and is meant to be as unobtrusive as possible. 

Downstream I am working on the logic to convert this image into something more local starting with the menus: 
https://github.com/linuxserver/docker-netbootxyz/tree/local

Outside of this PR I had some thoughts on how this could be achieved, but it is going to involve some pretty heavy re-writes of the ipxe files where the endpoints are defined along with templating I did not want to go off half cocked without gauging your interest. 
I have experience with all CM platforms including Ansible so if you want to start Jinja templating the files based on local/remote build parameters I am game.

Let me know what you think.

### DIFF
--- a/ipxe/disks/netboot.xyz
+++ b/ipxe/disks/netboot.xyz
@@ -5,6 +5,7 @@ set boldoff ${esc:string}[22m
 set fg_gre ${esc:string}[32m
 set fg_cya ${esc:string}[36m
 set fg_whi ${esc:string}[37m
+set TFTP_ERR Local TFTP failed... attempting remote HTTPS
 set HTTPS_ERR HTTPS appears to have failed... attempting HTTP
 set HTTP_ERR HTTP has failed, localbooting...
 set version 1.04
@@ -44,6 +45,7 @@ echo Attempting chainload of netboot.xyz...
 goto menu || goto failsafe
 
 :menu
+chain --autofree tftp://${next-server}/menu.ipxe || echo ${TFTP_ERR}
 set conn_type https
 chain --autofree https://boot.netboot.xyz/menu.ipxe || echo ${HTTPS_ERR}
 sleep 5


NiKiZe:
You will want to have this on http/https rather than tftp, it might also be a good idea to hijack a dhcp option, and only try local server boot if that option is set. That way there will be no delays

thelamer:
My thought process as it currently stands would be to have menu files initrd and kernels be on a tftp server distributed in the main image from Dockerhub. And have an optional env flag to enable an http/https server and pull down isos/recovery stuff to the users local machine via a bind mount. 
I am not familiar with DHCP boot options are there ones conventionally used for serving arbitrary metadata ?

antonym:
My future goal is to make netboot.xyz be able to be installed easily anywhere as a self hosted entity where you can either leverage internet mirrors for software bits or generate your own mirror of bits that you want to host.  What would be nice about generating your own mirrors is that you can then choose to crack open ISOs on your own and add many more options in your own environment.  I'd like those kind of recipes to be hosted in this repo to be used in local mirrors even though the netboot.xyz site might not host them.  We know how to boot a ton more ISOs but we're limited in the fact that they have to be cracked open first.

I was starting to go down the path of converting a lot of this to jinja templates in Ansible that would generate the entire menu set based on the configurations set.  You'd be able to disable menus, add new ones, etc by setting boolean flags.  That way the same code to generate netboot.xyz could be utilized anywhere else and then this repo might just turn into the code to generate the menus for the hosted site.  There's a lot of duplicated code in a lot of the scripts that could probably be condensed by using templates.  

In that case, it would also automatically generate the iPXE images along with the users desired certs and point to their own domain name, i.e. boot.user.com.  

But I'll have to give the docker image a whirl, I'd been wanting to do something like that but hadn't had the time to do it yet.  I'm open to all ideas on directions to take this.

thelamer:
@antonym 

I think I can start on some templating POC stuff, I will try to get some in this PR this week for review and discussion. (basic true/false ifs for displaying menu items, loops where it makes sense, and generic templates for user consumption by defining arrays in a custom yaml) We template all of our repos with Ansible right now https://github.com/linuxserver/docker-jenkins-builder so this is not a new concept for us.
From a core design point though do you think there is a hybrid mode here or is it all or nothing? 
My initial thought process revolved around looping through an array of required files to download locally to a server the user controls. I think that is **absolutely** needed for templated ipxe files, though the more I think about it hosting the files needed to boot really does not achieve anything if the installer/liveCD will need access to the internet and that same remote mirror for installation to continue, might as well grab the core files remotely also as you have already configured the project. I think simply having the boot.cfg localized to be able to set the mirrors to use is enough to satisfy the needs of self hosters to use their own mirrors along with templated customization directives so people don't get stuck on forked configs. 

As for cracking open ISOs and having network bootable stuff for more live CDs and Distros what do you think about spearheading that initiative ? Really boils down to where to host the squashFSs and people could just mirror off of your pipeline output ingesting the ISOs and spitting out the files to object storage. ( at our core as an org we maintain build server pipelines if that is needed )

I see you are involved with Rackspace and at Linuxserver.io we have a deal with Digital Ocean (they are very big on supporting open source stuff as an org) for their Spaces object storage offering. 

So the way I see it there are really 3 things here to expanding the functionality of the project: 
1. Add directives to the default bins to allow chain loading local config files without user re-compilation. 
2. Template the config files to a level where users can pass a local vars yaml to generate the config files on the fly even from rolling releases.
3. Setup build pipelines to upload ISO contents to object storage somewhere and document a process for mirroring those ( hopefully S3 compliant ) endpoints locally. 

I have always had a passion for polished PXE boot servers as a core infrastructure tool and have the time to dedicate to this kind of project.

thelamer:
@antonym 
Here is a basic loop for local templating of files and compilation: 
```
git clone --depth 1 https://github.com/thelamer/netboot.xyz.git -b templated
cd netboot.xyz/
docker build -t templater -f script/Dockerfile .
docker run --rm -it -e PUID=1000 -e PGID=1000 -v $(pwd):/buildout templater
```
The the output will be in menu and bootfiles. The templating is as minimal as possible with simple string substitution for the domain. This can be expanded to actual logic and loops as needed. 
I ripped out all of the code signing in the release build script as I honestly do not understand it. 

I thought a bit more about how to tear apart ISOs and host them for free and honestly I think Github might be perfect for this, release artifacts are essentially free object storage and can store files up to 2GB. Some basic travis logic can be setup to download the ISO extract the contents and upload the needed artifacts to a Github Release in their own repos. These releases as part of their build logic should then commit to this repo to update release endpoints for the menu with the new release SHA based on the sha256 of the ISO. 

I know I am stacking a lot on here but I have a few questions: 
1. Can this form of localization in this PR be merged ? (tftp try first)
2. Can you make a feature branch I can PR in logic for templating ?
3. Do you have an example ISO you have boot logic for but no HTTP endpoint? (would like to host and test in github)

thelamer:
@antonym I setup a POC for booting an Ubuntu LiveCD off of github here: 
https://github.com/thelamer/ubuntu-live-POC

You can boot with the following menu entry: 
```
set github_url https://github.com/thelamer/ubuntu-live-POC/releases/download/20191102035755-add4614b/
kernel ${github_url}vmlinuz boot=live config hooks=filesystem fetch=${github_url}filesystem.squashfs initrd=initrd
initrd ${github_url}initrd
```

The env is a little fucky as DNS is not set and I am not using casper, was hoping oyu might have a bit of insight on that.

antonym:
I think we may be overloading this PR up a bit.  It might be good to start tracking some of these ideas in an issue, so after this it might be good to open up issues and track them there so the community can view them easily or talk about them on Discord.

In regards to the PR, I'm a little hesitant to make a sweeping change to the primary bootdisks as it's hard to know what next-server or internal layout users are running, and a change to start polling for TFTP on boot might catch some by surprise.  

I think a better option would be to make a separate bootdisk that has localization built in and as we go down the path of getting a self option built, then we can communicate that behavior out and eventually merge it into the main code line.  I'm fine with generating additional disks as needed though.

We could also think about maybe adding a check for TFTP/localization that checks DHCP options and actions on them if they are set from the server.  That way, the hooks are in place, but don't impact the regular users.

I started going down the path of doing hosting on github for a few items here, but never really finished them.  I since learned how to do releases on github, so the artifacts could easily be kicked out as releases:

https://github.com/antonym/nixos-builder (using circle-ci)
https://github.com/antonym/gentoo-builder (using travis-ci)

Mainly they were small half-baked projects I started on and lost interest in as it's didn't seem like it was going to provide a ton of value or have a lot of interest.  But I thought it might be good to just press a button periodically or set up a watcher to go run it when it thought the release webpage might have changed.  Ultimately, I'd like to see a movement with the distros where they move away from using ISOs and move to a more portable installer kernel method.

I've been hacking on some Ansible the last two weeks that has pulled out the versions, mirrors, names out of the main ipxe templates into vars so that the src files are just jinja and it can generate the filetree.  It's really rough right now but it gets me to the point where i can generate the original source files.  I'll probably try and push up a WIP PR soon to start getting feedback or ideas.   The Docker branch looks interesting, I'd been wanting to do something like that for a while but just hadn't had the time.  I'll try and toy around with that soon.

thelamer:
I'll create issues and reference this PR. Also hop on discord to discuss kicking off the new templated/locally supported project we could ingest at Linuxserver. 
As far as an effort thing or these things being corner case usage scenarios I assure you there is a large audience out there just need the platform to reach them and the pipeline setup so users don't have to babysit the upstream changes. We setup some pipelines for us and we are near 100% automated only dealing with basic build issues https://github.com/linuxserver/pipeline-triggers . As for getting upstream to make changes I think if we build the pipelines and make common unobtrusive code mods that their users start being vocal about using as a consumption solution it will just naturally happen and we can stop hacking into images to apply patches. 

In any case I will close this PR.

PR #378, opened 2019-10-27
Update ALT Linux rescue URLs

danboid:
Update ALT Linux rescue URLs

### DIFF
--- a/src/utils.ipxe
+++ b/src/utils.ipxe
@@ -24,13 +24,13 @@ goto utils_exit
 
 :alt_linux_rescue
 iseq ${arch} x86_64 && goto alt_linux_rescue_x64 ||
-set util_path nightly.altlinux.org/p8/release/alt-p8-rescue-20180312-i586.iso
-set util_file alt-p8-rescue-20180312-i586.iso
+set util_path nightly.altlinux.org/p8/release/alt-p8-rescue-20190312-i586.iso
+set util_file alt-p8-rescue-20190312-i586.iso
 goto boot_memdisk
 
 :alt_linux_rescue_x64
-set util_path nightly.altlinux.org/p8/release/alt-p8-rescue-20180312-x86_64.iso
-set util_file alt-p8-rescue-20180312-x86_64.iso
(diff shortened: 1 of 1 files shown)


antonym:
Were you actually able to get it to load the rescue mode?  It pulls up the grub menu but then it searches for the mount drive once loaded into rescue.  More than likely it's not loading the memdisk drive into RAM properly.  Are there any other options useful other than the rescue that are needed?

danboid:
Hi Antony!

If I use my fork of the netboot.xyz-custom repo ( https://github.com/danboid/netboot.xyz-custom ) which has the updated URLs for the ALT Linux rescue ISOs, it starts to boot ALT but only gets as far as:

```
MEMDISK: bootstrap too large to load
```

I presume that error is because my netbook only has 1 GB RAM. I have not been able to test it on anything with more RAM yet. When using the current code/URLs for ALT it gives me an iPXE invalid URL error because the files it references don't exist any more.

I can still boot into Slackware but it is missing a few tools I would like that ALT will have such as rsync, p7zip, sgdisk, testdisk etc.

thelamer:
@danboid I just dug into this distro and it is Mandriva based their init hooks have and http option to load the squash so you can use something like this: 

```
:CUSTOM
set server http://192.168.10.13
kernel ${server}/vmlinuz live fastboot automatic=method:http,network:dhcp,server:192.168.10.13,directory:/ stagename=live ramdisk_size=904173
initrd ${server}/full.cz
boot
```

The fundamental problem with their init hooks is that it is a pre-compiled binary `/sbin/init-bin` and given the argument structure it does not support https endpoints (where we host assets) or specifying a port for the webserver. 
I tried to dig into their source and see if it is something we could patch but honestly their is a language and community barrier here:
https://www.altlinux.org/Gear/remotes

In any case we are migrating away from ever using memdisk to load raw ISO in general as a project so this would not make it into our main set of menus for the 2.x series. So this will need to stay in your custom repo. 

If any of this changes down the line and their project can support pulling in squash files from https sources and specify webserver ports feel free to open a new issue we can host their ISO assets and provide boot options for their Distro.

PR #382, opened 2019-11-04
nixos.ipxe: fix version mistake

ghost:
Selecting the NixOS 19.09 entry was booting 19.03 because of a mistake


antonym (approved):
Good catch, thanks!

PR #386, opened 2019-11-07
Updated Clonezilla to 2.6.4-10 and switched to osdn.net mirror.

samip5:


antonym:
Any particular reason for the mirror change?  I'd probably keep the mirror as is, but we can bump the version.

samip5:
@antonym The mirror change was due to awfully slow-loading time from the other mirror and to my knowledge, osdn usually works the best, due to their geo-download handling if I'm not mistaken.

thelamer:
@samip5 the problem with OSDN even though they have better mirrors is their URL format. It redirects to mirrors with unique subfolders IE https://mirrors.gigenet.com/OSDN/clonezilla/65892/ makes it hard to just change version numbers. 

Sorry for all the commits this one should be squashed.

thelamer (approved):

antonym:
Might want to leave it as master.dl.sourceforge.net as I had better success with that URL as being stable.  The downloads URL seems to be unstable and iPXE doesn't like the redirects too much.

samip5:
The problem that at least in Europe I came across is that master.dl.sourceforge.net is REALLY slow and OSDN works much better and it worked okey with the redirects at least when I tested it.

thelamer:
Yeah , it looks like until we start migrating assets to internal GitHub releases for publishing and in turn giving users ability to self host via their own http servers we should avoid SF redirects . (Use master)
As for the OSDN mirror again I agree with you on speed it really boils down to their asset structure , the URL will need to updated with the version for every change due to them not being generic.

antonym:
+1... really anything we attempt to load as an ISO will probably be a candidate for publishing on Github Releases as we can't really load them in EFI mode and memdisk has a lot of limitations.  Legacy mode will start to be phased out on newer hardware and we'll really want these tools to be flexible in both types of modes in a more efficient way than loading ISOs into RAM.  I get that speed is an issue, but I'm favoring stability and maintainability all over that for now until we can get a better more reliable way of geo load-balancing files.

samip5:
@thelamer I squashed your commits to one.

antonym:
Updated in development, closing this out.

PR #389, opened 2019-11-15
Adds experimental Live CD menus for Debian and Ubuntu

antonym:
Adds new extracted and hosted CDs for booting Debian
and Ubuntu Live up into memory.  Support is experimental
and work is ongoing.


thelamer:
If it works the Debian kernel version should be updated on this to support the multi part ones

thelamer:
You should be able to regen to get the latest updates and have a working set across ubuntu and Debian. 
Did the 19.10 boots not need altered entries for url or fetch ?

antonym:
I tried the 19.10 Kylin, XFCE, LXDE, etc and got this when loading:

image

I ran across https://bugs.launchpad.net/ubuntu/+source/casper/+bug/1848018 and added ip=dhcp but but it didn't seem to help.

antonym:
Ah, never mind, I see what you meant, have to use netboot=url and url= for 19.10.

antonym:
So it gets farther after the syntax changes but fails when trying Kylin:

Screen Shot 2019-11-16 at 7 49 45 AM

antonym:
19.10 Mate does work now though, so it may be that the Kylin image needs the multipart stuff as well.

thelamer:
I was worried about the tmpfs sizing args in Casper , I honestly thought they were not working and it just mounted half of RAM available. 
Will dig into it tommorow

thelamer:
> So it gets farther after the syntax changes but fails when trying Kylin:
> 
> Screen Shot 2019-11-16 at 7 49 45 AM

@antonym does the VM you are doing loop tests in only have 4 gigs of memory allocated to it by chance? The content length thing should not be working (it is in there for shit that is not our weird S3 based endpoint) so the tmpfs should be mounted with half of the systems memory which would explain why a system with a 2 gig tmpfs space would not work with the bigger images.

antonym:
Yeah, originally I tested with 4GB but I got the same with 8Gb of RAM in Virt Box with Kylin, the 19.10 Budgie also doesn't even seem to try and download, but just goes to can't find live filesystem.

thelamer:
budgie is due to the wrong kernel being set in settings it is a small squash 1.5 gigs. (new release with proper endpoints for that https://github.com/netbootxyz/netboot.xyz/commit/991233ae6ad885811e0ade2086d57b7740566897)

Kylin is the only multiparter , the only thing that looks off is the squash error before the thing finishes downloading. It is like the logic continues even while the download is still going, maybe a wait command in there ?

PR #398, opened 2019-12-04
Updated Archlinux to 2019.12.01

Centzilius:


thelamer (approved):

thelamer:
@antonym 

Only question I have is should we include all 3 months in the menu or just merge this as is ? 
http://mirror.rackspace.com/archlinux/iso/

antonym:
I think latest should be sufficient enough as usually it's going to be someone trying a new install and will want the latest anyways.

I'm working through some sig stuff so you may have to disable sigs for now on this one.

PR #401, opened 2019-12-06
bump to 5.1

forkiesassds:


thelamer (approved):

thelamer:
Actually went with a more generic 5 definition in the main repo , but you are spot on with the asset endpoint change.

PR #402, opened 2019-12-06
add support for q4os live

forkiesassds:


thelamer:
This would also need the live.ipxe entry to link to it

forkiesassds:
I added the live.ipxe entry.

src/live.ipxe
```diff
@@ -9,6 +9,10 @@ item --gap Live Boot Distributions
 item live-debian ${space} Debian Live
 item live-fedora ${space} Fedora Live
 item live-ubuntu ${space} Ubuntu Live
+item live-kali ${space} Kali Live
+item live-tails ${space} Tails Live
+item žlive-elementary ${space} elementary OS Live
```
antonym: Bad characters

antonym (changes requested):
Bad characters

src/live.ipxe
```diff
@@ -9,6 +9,10 @@ item --gap Live Boot Distributions
 item live-debian ${space} Debian Live
 item live-fedora ${space} Fedora Live
 item live-ubuntu ${space} Ubuntu Live
+item live-kali ${space} Kali Live
```
thelamer, line 12: These items have been in alphabetical order should be like: 
live-debian
live-elementary
live-fedora
live-kali
live-q4os
live-tails
live-ubuntu

Also Please drop Tails, as is it should only be in development releases until we can figure out how to get networking working.

thelamer (changes requested):

src/live.ipxe
forkiesassds: There fixed.

src/live.ipxe
forkiesassds: There. I have sorted them by alphabetical order, and dropped Tails.

thelamer (approved):

antonym:
Pushing a generated dump from development up for testing here: https://github.com/antonym/netboot.xyz/pull/403

Closing this one out for that one, which should also cover Q4OS

PR #414, opened 2019-12-12
Add news to menu

antonym:
Attemps to load project news from project site, and if unavailable
loads the news from within the checked out release


thelamer (changes requested):
Should we use the site name instead of hardcode to boot ? to allow people to do custom news and for us to stage this to see live ?

antonym:
I kind of see it as the "About netboot.xyz", the overall project page... maybe news is the wrong way to frame it, but figured there should be something that links it back to the original project regardless if it's self-hosted or customized.  It might be good to incorporate the current checked out versions, and maybe a version check page as well to let the user know there's a newer version available.

thelamer (approved):

thelamer:
Fair enough

PR #423, opened 2019-12-15
Fixing signatures and add certs to loaders

antonym:
* Fixing some index code
* Shifting after_deploy messages and aws invalidation to
  proper locations
* Add cert checkout to preinstall and setting those on s3
  deploys only
* Adding production docker file which runs netboot.xyz overrides
* Only display sig menus if enabled
* Fix var names for signing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,10 @@
 sudo: true
-
 language: bash
-
 services:
   - docker
-
 env:
   global:
     - DEBIAN_FRONTEND="noninteractive"
-
 jobs:
   include:
     - stage: test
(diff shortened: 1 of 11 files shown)


thelamer:
The thing to keep in mind about the before deploy and after deploy is they run before and after every single deploy in that stage.

So because we only want to tag once and invalidate once and install aws cli once we do that in the GitHub deploy step. 
T

antonym:
Ok, ran through the code again when I wasn't half asleep and that makes sense.  I created another script to run through the cert retrieval on the deploy to s3 stages and left the others as is.

PR #436, opened 2019-12-21
forgot lz file extension

thelamer:


antonym:
You'll need the initrd=initrd.lz on the kernel line for the EFI stub otherwise it'll fail on EFI mode.

antonym (changes requested):
You'll need the initrd=initrd.lz on the kernel line for the EFI stub otherwise it'll fail on EFI mode.

antonym (approved):

PR #455, opened 2019-12-27
fix ambiguous param in geniso image

vincentaubert:
Hello

while using ansible to generate my own netboot.xyz instance on a debian 10 server
I had this error : 
genisoimage: option '-e' is ambiguous; possibilities: '-eltorito-boot' '-eltorito-alt-boot' '-eltorito-catalog' '-exclude' '-exclude-list' '-ethershare' '-exchange'

I fixed it simply with this code in the PR


antonym:
I ran across this when I did a few Debian tests and hadn’t circled back to it yet.  If we switch to this flag, I think it then breaks CentOS and Ubuntu.  What I’ll probably do is just make the parameter or the command a OS specific var to get around that issue.

antonym:
Looks like this should work for the others, I guess they made a change from -e to -eltorito-boot at one point in the tool where it was inconsistent.

antonym (approved):

PR #462, opened 2019-12-29
Windows Testing Fixes

thelamer:
I wanted to automate Windows download and file org stuff in the web app , went pretty far down the rabbit hole. 
First off good information here: 
http://mistyrebootfiles.altervista.org/documents/TinyPXEServer/files/winpe_wimboot.htm

Without imgfree and the sig check stuff removed I always got `bad cpio magic` from wimboot when trying to extract the files. 

I got this booting in VM and bare metal, Bare metal gets farther in the install process though it asks for a USB/CD containing drivers to load up stuff. 
To do win7 all the files would need to be renamed, though given it is EOL I think we should be good with win10 support only.


antonym:
Sounds good, I'm also good with Windows 10 support to start as well.

PR #467, opened 2019-12-31
Release 2.0.2

antonym:
Contains Fixes for Windows, timeout for version checking, and several fixes for Deepin and Elementary.


thelamer:
Found a bug in deepin

antonym:
K, I’ll hold off.

thelamer:
We would need to version bump again unless we don't care about pushing a new RC

antonym:
I had a change to slip in too so just cherry-picked them into RC and we can go with those without a version change.

Issue #479, opened 2020-01-02
Add Support for FreeBSD


antonym:
Get FreeBSD working either with iPXE or with FreeBSD assets.  The existing solution was using memdisk to load a raw hard disk image and broke Legacy when the cert used for mfsbsd no longer worked with https.

https://github.com/netbootxyz/netboot.xyz/issues/302

We'd probably want this working with UEFI if possible too so we'd want to move away from the memdisk solution.

thelamer:
You wanna keep this open trying to function with UEFI?

antonym:
We can always revisit later, at least base functionality is working again but over the next few years I think Intel plans on dropping legacy support, so we should always try and favor UEFI when possible. Thanks for knocking this out!

PR #480, opened 2020-01-03
#479 enabling freebsd and using mirrored assets to boot

thelamer:
Pull the menu from endpoints so we don't need to mess with the templates unless we want to. Tested booting 8 and 11 default and SE. 

ref: #479


antonym (approved):
Nice, looks good!

PR #483, opened 2020-01-03
Save the ipxe_version compiled version

antonym:
Saves the original version variable that contains the ipxe version
as ipxe_version so that it can be referenced as needed.

Also ensures the other ipxe disks are using the correct boot_version
variable.


thelamer:
Does it work in this order?

antonym:
Yeah, I think so but I’ll test in dev. IPXE by default sets that var to the iPXE version.  I could rename all vars in everything but this seems easier if it works. Then I can add the upstream iPXE version to about.ipxe since it looks like they are going to start incrementing versions upstream.

https://github.com/ipxe/ipxe/commit/8f1514a00450119b04b08642c55aa674bdf5a4ef

antonym:
Works, this captures the ipxe version now:

Screen Shot 2020-01-03 at 10 59 37 AM

Issue #219, opened 2018-02-01
VLAN tagged interfaces


sumkincpp:
Would be great to be able to specify vlan tag(802.1q) for networking configuration during initial boot.

willglynn:
One option: use [ROM-o-matic](https://rom-o-matic.eu) to make a custom iPXE image with an embedded script, [`vcreate`](https://ipxe.org/cmd/vcreate) your VLAN, configure the interface, and [chain into netboot.xyz](https://netboot.xyz/booting/ipxe/).

Would you be wanting to configure this interactively at boot time?

thelamer:
This will be in the next release with https://github.com/netbootxyz/netboot.xyz/commit/67ab2b4729c9517d222a058d12e5b7b083703a3a

If you want to test it now you can pull a head build: 
https://s3.amazonaws.com/dev.boot.netboot.xyz/7047aeb500e5edbcced514d953583d84043e1851/index.html

PR #487, opened 2020-01-04
adding vlan manual configuration to boot disk

thelamer:
I cannot full loop this as I lack the network infra. 
The logic and menu interaction checks out though. 

Ref: https://github.com/netbootxyz/netboot.xyz/issues/219


antonym:
Nice, will try and test this in my lab, think I should be able to simulate it. Will merge to get the builds rolling.

antonym (approved):

Issue #485, opened 2020-01-04
FerenOS Live support


thelamer:
https://ferenos.weebly.com/get-feren-os.html Ubuntu 18.04 based pretty straight forward to make the live assets. 
They do wine and drivers out of the box, and nice software choices.

PR #488, opened 2020-01-04
adding ferenOS to live menu

thelamer:
Standard does not play nice with Virtualbox but is good on bare metal. 

Standard live boot add here, manual babysit version until they have some kind of mirror off sourceforge with versioning setup.


thelamer:
ref: https://github.com/netbootxyz/netboot.xyz/issues/485

antonym (approved):

PR #490, opened 2020-01-06
Clonezilla live assets, support live assets in utility menu

thelamer:
In order to use identical key names for menu entries we need a separate yaml array for UEFI stuff. 

This does not offer any scripted customization from the ipxe menu though it should be OK for an initial release it is certainly better than memdisk. 
Also on boot Clonezilla lets you set remote scripted install locations so not sure what benefit there is to doing it in IPXE. If a user want to use hooks they should be local hosting and customizing the clonezilla.ipxe file.

### DIFF
--- a/roles/netbootxyz/templates/menu/utils-efi.ipxe.j2
+++ b/roles/netbootxyz/templates/menu/utils-efi.ipxe.j2
@@ -2,8 +2,8 @@
 
 menu Utilities - Image Sig Checks: [${img_sigs_enabled}]
 item --gap Utilities:
-{% for key, value in utilities.items() | sort(attribute='1.name') %}
-{% if value.enabled | bool and value.menu == "efi" %}
+{% for key, value in utilitiesefi.items() | sort(attribute='1.name') %}
+{% if value.enabled %}
 item {{ key }} ${space} {{ value.name }}
 {% endif %}
 {% endfor %}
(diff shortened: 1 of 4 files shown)


roles/netbootxyz/templates/menu/utils-efi.ipxe.j2
```diff
@@ -17,8 +17,8 @@ goto ${menu} ||
 chain ${menu}.ipxe || goto utils_exit
 goto utils_exit
 
-{% for key, value in utilities.items() | sort %}
-{% if value.enabled | bool and value.menu == "efi" %}
+{% for key, value in utilitiesefi.items() | sort %}
+{% if value.enabled | bool and value.type == "memdisk" %}
```
antonym, line 21: Probably won't need the memdisk type check if it's EFI.

antonym (changes requested):

roles/netbootxyz/templates/menu/utils-efi.ipxe.j2
thelamer: That logic specifically guards the memdisk menu entries and  allows the core one to boot from the clonezilla.ipxe file.

antonym: Yeah, just saying we can probably remove all of the memdisk stuff from utils-efi.ipxe.j2 while we're at it because it won't ever work in efi mode.

thelamer: I concur you want it in the scope of this PR?

antonym: I'll merge this one in, and we can clean up in another.

antonym (approved):

PR #493, opened 2020-01-06
add DBAN from assets for UEFI and update README

thelamer:
Pulls the kernel from our asset mirror, I duped their current boot options in IPXE and tested functionality. 
User is forced through a warning screen with a default back option.


antonym (approved):

patrickbussmann:
I dont know why but when I download the newest version and I select "Utilities" then `DBAN` is not visible there.
Maybe its deleted in another PR.

https://github.com/netbootxyz/netboot.xyz/pull/505 here is DBAN deleted from a file.
https://github.com/netbootxyz/netboot.xyz/pull/497 here it was set from enabled to disabled

Hmm

PR #503, opened 2020-01-09
Enable text installs on RHEL and Fedora

antonym:
Adds options for text installs on RHEL/Fedora.


thelamer:
I don't have rhel assets to test but the flow if this does not make sense, does the `rhel_install_text` item need a goto statement ?

antonym:
It should flow to the next part rhel_install.  I could put a goto rhel_install too, it does the same thing as it's the next line in the sequence.

antonym:
Added the gotos just to make it read a bit better.

thelamer (approved):

Issue #494, opened 2020-01-06
Add support for K3OS


OrvilleQ:
Is there any plan for K3OS support?

It should be similar to Rancher OS.

https://github.com/rancher/k3os

antonym:
Sure, we'll look into it and add it if viable.

https://github.com/rancher/k3os/releases

thelamer:
ref: https://github.com/rancher/k3os/issues/56

antonym:
Iniital support merged into development: https://github.com/netbootxyz/netboot.xyz/pull/505

PR #505, opened 2020-01-09
#494 add k3os, remove non functional uefi apps, fix dban naming

thelamer:
Pretty basic here, there is a ISO warning on boot for 5 ticks but it does not effect functionality. 

The installer lets you set boot time settings do not see a point in embedding that in IPXE menu stuff. 

Also went live vs install as the install can be started from live with the default command echoed in the MOTD.


antonym (approved):
LGTM

PR #506, opened 2020-01-10
Adds netboot.xyz endpoints menu for testing

antonym:
Uploads version.ipxe to root of endpoints so that latest
deployed version can be loaded.

Removes memdisk iso test from efi utils menu.


antonym:
Danke!

PR #510, opened 2020-01-10
enable auto login for rancher and prune uneeded options

thelamer:
They have auto login option now among other new stuff: 

https://github.com/rancher/os/releases/download/v1.5.5/rancheros.ipxe

From a live web bootable standpoint this is the only one we want though.


antonym (approved):
Cool, lgtm

PR #511, opened 2020-01-11
Sparkylive

thelamer:
This puts us in scrolling territory, wish ipxe did columns we have plenty of X space. 

Tested the boot entries locally.


antonym (approved):
Guess we could always add another menu of groups alphabetically if we needed to (a-m)(n-z).

PR #522, opened 2020-01-15
adding devuan linux to live menu

thelamer:
Debian fork with no systemD and running on patched up gnome2. It is Jessie based but they are keeping up to date and patching stuff. 

Tested hardware/VM UEFI/PCbios

### DIFF
--- a/roles/netbootxyz/templates/menu/live-devuan.ipxe.j2
+++ b/roles/netbootxyz/templates/menu/live-devuan.ipxe.j2
@@ -1,0 +1,37 @@
+#!ipxe
+
+goto ${menu} ||
+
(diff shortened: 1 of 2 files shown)


antonym:
Kind of wonder how much it's being maintained, I had the regular images disabled because the mirror I was using had a cert that was a year out of date:

http://auto.mirror.devuan.org

thelamer:
It seems like they know the main mirror url has been down since November it is noted in the mirror list:
https://pkgmaster.devuan.org/mirror_list.txt

The last point release of the ascii build was September 2019 .

antonym (approved):

PR #523, opened 2020-01-15
update boot.wim path

drizuid:
boot.wim is generally located in the sources directory of the image. The "echo" showed the correct path, but the initrd command attempted to load from the wrong path.


antonym (approved):

antonym:
Good catch, thanks!

PR #524, opened 2020-01-15
Create WindowsPE.ipxe.j2

drizuid:
Windows PE has some case differences from windows ISOs in that the boot folder is "Boot" and the bcd file is BCD.  This was also briefly discussed in discord.


drizuid:
i didn't modify the menu in this commit, so it is not available in the distributions list

PR #525, opened 2020-01-15
Additions to support WindowsPE

drizuid:
Added a windows menu to select between a Windows Install disc and the Windows Presinstallation Environment. This was necessary due to some case differences between windows images and PE, namely that in the windows image, the boot directory is "boot" and bcd is "bcd" but in PE, boot is "Boot" and bcd is "BCD"  While this doesn't matter in windows, it does in a linux based boot environment and caused a failure. 

I tested this on my personal machine.

### DIFF
--- a/roles/netbootxyz/templates/menu/menu.ipxe.j2
+++ b/roles/netbootxyz/templates/menu/menu.ipxe.j2
@@ -50,7 +50,7 @@ iseq ${menu_live} 1 && item live ${space} Live CDs ||
 iseq ${menu_bsd} 1 && item bsd ${space} BSD Installs ||
 iseq ${menu_freedos} 1 && item freedos ${space} FreeDOS || 
 iseq ${menu_security} 1 && item security ${space} Security Related ||
-iseq ${menu_windows} 1 && item windows ${space} Windows ||
+iseq ${menu_windows} 1 && item winmenu ${space} Windows Installs ||
 item --gap Tools:
 iseq ${menu_utils} 1 && item utils-${platform} ${space} Utilities ||
 iseq ${arch} x86_64 && set bits 64 || set bits 32
--- a/roles/netbootxyz/templates/menu/winmenu.ipxe
+++ b/roles/netbootxyz/templates/menu/winmenu.ipxe
@@ -1,0 +1,26 @@
+#!ipxe
+
+goto ${menu} ||
+
+:winmenu
+menu Windows Installers - Current Arch [ ${arch} ]
+iseq ${arch} x86_64 && set arch_a amd64 || set arch_a ${arch}
+item --gap Windows Installers:
+item winpe ${space} Windows Preinstallation Environment
+item windows ${space} Windows Images
+choose menu || goto windows_exit
+echo ${cls}
+goto ${menu} ||
+iseq ${sigs_enabled} true && goto verify_sigs || goto change_menu
+
+:verify_sigs
+imgverify ${menu}.ipxe ${sigs}${menu}.ipxe.sig || goto error
+goto change_menu
+
+:change_menu
+chain ${menu}.ipxe || goto error
+goto winmenu
+
+:windows_exit
+clear menu
+exit 0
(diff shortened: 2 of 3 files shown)


thelamer:
This does not need a menu and a separate file you could achieve this with a variable and menu entry like this: 
```
#!ipxe

# Microsoft Windows
# https://www.microsoft.com

set win_arch x64
goto ${menu} ||

:windows
set os Microsoft Windows
clear win_version
menu ${os}
set bcd /boot/bcd
set sdi /boot/boot.sdi 
item --gap Installers
item win_install ${space} Load ${os} Installer...
item win_pe ${space} Load ${os} Preinstallation Environment...
item --gap Options:
item arch_set ${space} Architecture [ ${win_arch} ]
item url_set ${space} Base URL [ ${win_base_url} ]
choose win_version || goto windows_exit
goto ${win_version}

:arch_set
iseq ${win_arch} x64 && set win_arch x86 || set win_arch x64
goto windows

:url_set
echo Set the HTTP URL of an extracted Windows ISO without the trailing slash:
echo e.g. http://www.mydomain.com/windows
echo
echo -n URL: ${} && read win_base_url
echo
echo netboot.xyz will attempt to load the following files:
echo ${win_base_url}/${win_arch}/bootmgr
echo ${win_base_url}/${win_arch}/bootmgr.efi
echo ${win_base_url}/${win_arch}${bcd}
echo ${win_base_url}/${win_arch}${sdi}
echo ${win_base_url}/${win_arch}/sources/boot.wim
echo
prompt Press any key to return to Windows Menu...
goto windows

:win_install
isset ${win_base_url} && goto boot || echo URL not set... && goto url_set

:win_pe
set bcd /Boot/BCD
set sdi /Boot/boot.sdi
isset ${win_base_url} && goto boot || echo URL not set... && goto url_set

:boot
imgfree
kernel {{ wimboot_location }}
initrd -n bootmgr     ${win_base_url}/${win_arch}/bootmgr          bootmgr ||
initrd -n bootmgr.efi ${win_base_url}/${win_arch}/bootmgr.efi      bootmgr.efi ||      
initrd -n bcd         ${win_base_url}/${win_arch}${bcd}            bcd
initrd -n boot.sdi    ${win_base_url}/${win_arch}${sdi}            boot.sdi   
initrd -n boot.wim    ${win_base_url}/${win_arch}/sources/boot.wim boot.wim
boot

:windows_exit
exit 0
```

drizuid:
Ah much more elegant than what i was trying to do. I knew there had to be a way, but i couldn't envision it, thanks!
the only issue i see there, and it's very minor is that the echo statements will be wrong because you have to set the URL prior to selecting PE vs non-PE. 

the only other note i have is that in testing this, it didn't work. setting the base URL to my winpe or win10 and selecting the corresponding option resulted in some errors; changing 
`kernel {{ wimboot_location }}` to `kernel http://${boot_domain}/wimboot` resolved the issue. I'll leave this up so i can finish testing tomorrow.

antonym:
We could also just keep the existing option and attempt various variations of loading the files:

initrd -n bcd         ${win_base_url}/${win_arch}/boot/bcd            bcd ||
initrd -n bcd         ${win_base_url}/${win_arch}/Boot/BCD          bcd ||
initrd -n boot.sdi    ${win_base_url}/${win_arch}/boot/boot.sdi            boot.sdi ||
initrd -n boot.sdi    ${win_base_url}/${win_arch}/Boot/boot.sdi           boot.sdi ||

If it fails one, it'll just reattempt the next.  I definitely think we can probably do this with a single menu option though.

drizuid:
I've tested both of the proposed options and both work. I imagine the OR is the more seamless option as it requires no extra effort from the end-user, though this does still leave the echo commands incorrect which I think is only a minor issue. I'll close this PR as it sounds like the 2 proposals are much better than my ghastly change :D

my only change to ryan's was `kernel {{ wimboot_location }}` to `kernel http://${boot_domain}/wimboot`

i tested Antony's with no changes, both options work perfectly.

PR #527, opened 2020-01-16
adding parrot live modify current entry into an installer

thelamer:
We might want to think about dropping the security related menu or conversely shifting all the liveCDs and installers that are security/privacy focused into it. For this one I moved parrot to the linux installers menu. 

The Live stuff has the Vbox bug where you need to mash buttons on the initial curl, but works fine on hardware. 

Tested on virtual and hardware Bios/UEFI.


antonym:
Yeah I'm good with dropping the security menu and merging, should free up some space.

antonym (approved):

Issue #528, opened 2020-01-16
Add support back for grml


antonym:
Add support back for grml, we used to use memdisk and memdiskfind to load, but we can probably switch to hosting the assets so that maybe we can support EFI too

Original reference here:
https://github.com/netbootxyz/netboot.xyz/issues/70

PR #529, opened 2020-01-17
#528 adding grml menu and assets for live booting

thelamer:
I only tested this in VM but it is a modern kernel should work in UEFI. 

Left the live entry in the default yml as is.


antonym (approved):
LGTM, just needs a rebase.

PR #538, opened 2020-01-18
add velt live OS, and add coreos to readme

thelamer:
Unique arch spin with Velt UI.


antonym (approved):

antonym:
Thanks for the doc update!

PR #541, opened 2020-01-18
adding anarchy installer from asssets

thelamer:
Basic text based installer livecd. Tested in VM.


roles/netbootxyz/defaults/main.yml
```diff
@@ -84,6 +84,13 @@ releases:
         code_name: "v3.11"
       - name: "Edge (development)"
         code_name: "edge"
+  anarchy:
+    name: "Anarchy Linux"
+    enabled: true
+    menu: "linux"
+    versions:
+      - name: "current"
+      - code_name: "current"
```
antonym: Think it's failing on the second hyphen on code_name.

antonym (changes requested):

PR #546, opened 2020-01-18
add ip params to arch variants

thelamer:
This will cut a new RC when merged up, after this is merged we can change the version back to 2.0.6 and merge up again. 

The RC build will fail but we can then release under the 2.0.6 tag.


thelamer:
hold up looks like zen is broken need to fix that

antonym (approved):

PR #560, opened 2020-01-25
Allow more than one RC at given version

thelamer:
I went with the entropy approach here, simplifies the logic to just checking for if it exists instead of drilling into paginated APIs in a loop. The chances for collision are pretty slim unless we push RCs with 100s of revisions. 
In the unlikely events that we need to push multiple RCs this will generate new tags if the original -RC release already exists. 

Also we should cut an RC today.


antonym (approved):
The approach looks good, will try and get an RC out today.

PR #576, opened 2020-02-14
Release 2.0.9-RC

antonym:
Various version bumps


thelamer:
I have tested Ubuntu sparky and Debian point releases, most flavors.

PR #587, opened 2020-03-28
Fix utilities override

Enrico204:
This pull request fixes the override of the "utilities" section, as it was splitted in "EFI" and "PCBIOS" in `main.yml` (`utilitiesefi` and `utilitiespcbios`) but not in `user_overrides.yml` documentation, nor in `tasks/generate_menus.yml` (Ansible role `netbootxyz`).

Also, by following the documentation in `user_overrides.yml` comments (that says to create a `utilities_override:` section) we incur in an error during the Ansible role execution of the relative task (as `utilities` is now undefined).

This pull request fixes the role and update the documentation in `user_overrides.yml`

### DIFF
--- a/roles/netbootxyz/tasks/generate_menus.yml
+++ b/roles/netbootxyz/tasks/generate_menus.yml
@@ -9,15 +9,27 @@
       releases: "{{ _releases }}"
     when: release_overrides is defined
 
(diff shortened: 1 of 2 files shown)


antonym (approved):
Thanks! LGTM

PR #616, opened 2020-04-27
Bump display name ManJaro

thelamer:
Depending on the next release will likely need to remove the old installer concept and just keep the rolling current one dropping the labels.


antonym:
We need to bump old stable to 19.x?

antonym (approved):

thelamer:
Well that's the thing 19.x only lived for like a month or two, 18.x was for a year or more and saw a lot of penetration . 
We would optimally kill off the oldstable part come next release and just have their current jnsfaller as it is a rolling distro.

antonym:
Cool, sounds good!

PR #618, opened 2020-04-29
add fedora 32 live assets and bump version in prep for release

thelamer:
Maintenance, I would think we will likely do the three in a loop and EOL 31 purging assets when 34 is out. Not really a great way to automate this so will be manual for now.


antonym (approved):
LGTM

PR #620, opened 2020-05-01
adding focal images for PopOS

thelamer:
Untested, thought will comment when I get a minute to loop, figured I would open this before I pass out.


thelamer:
Tested Functional

antonym (approved):

Issue #623, opened 2020-05-02
Ansible playbook does not build on Fedora


rwaltr:
**Describe the bug**
The Ansible playbook `generate_disks_base` attempts to install EPEL on all Redhat OS family distributions, Fedora Linux and Fedora containers reports on Ansible that they are "Redhat" on ansible_os_family, Because Fedora does not include the EPEL repository and matches the contitions for task "Ensure EPEL is enabled", This playbook will fail.

**To Reproduce**
1. Create a Fedora container, One quick way of doing so is ``[podman/docker] run --rm -it fedora``
1. Clone the Netboot repo and install ansible, git, and httpd
1. Run the playbook using `ansible-playbook

**Screenshot**
```ansible
TASK [netbootxyz : Ensure EPEL is enabled] *************************************
fatal: [localhost]: FAILED! => {"changed": false, "failures": ["epel-release No match for argument: epel-release"], "msg": "Failed to install some of the specified packages", "rc": 1, "results": []}
````

**Expected behavior**
EPEL would be skipped on Fedora Linux

**Additional context**
The variable `ansible_distribution` will show `"Fedora"` `"CentOS"` and would be more reliable 

I'm currently looking at a workaround now

rwaltr:
I have created pull #625 to resolve this

antonym:
Thanks, this has been merged.

PR #624, opened 2020-05-02
WIP: Enable running playbook on Fedora

rwaltr:
Fedora includes most EPEL packages inside of their default repo, they also do not have a epel package,
including this "and not" statment will exlude Fedora from attempting to install epel
This is for #623


rwaltr:
Looking farther into this, It appears that more than  just adding an exclusion will be required to let this build on fedora. Will look into this further

PR #632, opened 2020-05-10
Get CentOS 7 working correctly.

### LINKED ISSUE #631: Can't install CentOS
Pretty sure this is not a bug from netboot.xyz. There is no BASEOS folder in http://mirror.centos.org/centos/7.8.2003/ for CentOS 7.8.2003 so CentOS can't be installed via netboot.xyz

samip5:
Using the right endpoint as instructed by their README in the mirror.
Might break something.

Fixes #631

### DIFF
--- a/roles/netbootxyz/defaults/main.yml
+++ b/roles/netbootxyz/defaults/main.yml
@@ -127,8 +127,8 @@ releases:
       name: '8.0'
     - code_name: 8-stream
       name: 8.0 Stream
-    - code_name: 7.8.2003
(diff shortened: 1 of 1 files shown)


samip5:
@antonym Please review this and maybe merge. I don't know if it will break something but I don't think so?

antonym:
Thanks for the PR, I had to make a few other adjustments and put them up here:

https://github.com/netbootxyz/netboot.xyz/pull/633/files

PR #651, opened 2020-06-09
Improved k3os install and added live-boot

pvelati:
Added customizable parameters (see image attached below):
image 1
- version of k3os to install (defaults to latest, taken directly from github, no update needed for latest version, it will be always latest, using release tags. Specific version can also be defined, for example v0.10.2 or v0.10.1) (see image attached below).

- install device (defaults to /dev/sda)
- config.yaml url (empty as default): url for the installation yaml (see [sample-config](https://github.com/rancher/k3os#sample-configyaml) ) 
- mirror url (defaults to github releases page): can be modified if you self-host the files (note: the structure must match the same used by github releases page)

A "custom" installation example can be viewed in image below.

A live-boot version of K3OS is also added to live-boot menu (some parameters can be modified before booting).

### DIFF
--- a/roles/netbootxyz/templates/menu/k3os.ipxe.j2
+++ b/roles/netbootxyz/templates/menu/k3os.ipxe.j2
@@ -7,7 +7,8 @@
 goto ${menu} ||
 
 :k3os
-set os k3OS
+set os {{ releases.k3os.name }}
+set k3os_mirror {{ releases.k3os.mirror }}
 isset ${k3os_version} || set k3os_version latest
 isset ${k3os_install_device} || set k3os_install_device /dev/sda
 menu ${os} by Rancher - ${arch_a}
--- a/roles/netbootxyz/templates/menu/k3os.ipxe.j2
+++ b/roles/netbootxyz/templates/menu/k3os.ipxe.j2
@@ -1,24 +1,76 @@
 #!ipxe
 
-# k3OS
+# k3OS by Rancher
 # https://k3os.io/
+# https://github.com/rancher/k3os#configuration-reference
 
 goto ${menu} ||
 
 :k3os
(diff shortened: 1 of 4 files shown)


roles/netbootxyz/templates/menu/k3os.ipxe.j2
```diff
@@ -1,24 +1,76 @@
...
-# k3OS
+# k3OS by Rancher
 # https://k3os.io/
+# https://github.com/rancher/k3os#configuration-reference
 
 goto ${menu} ||
 
 :k3os
-set os {{ releases.k3os.name }}
-set k3os_mirror {{ releases.k3os.mirror }}
```
antonym, line 10: You'll still need to set this mirror variable once as it's pulling it from the defaults above.

roles/netbootxyz/templates/menu/k3os.ipxe.j2
```diff
@@ -1,24 +1,76 @@
...
 
-# k3OS
+# k3OS by Rancher
 # https://k3os.io/
+# https://github.com/rancher/k3os#configuration-reference
 
 goto ${menu} ||
 
 :k3os
-set os {{ releases.k3os.name }}
```
antonym, line 9: Needs to be set as it pulls it from the defaults yaml in case there are future changes.

antonym (changes requested):
Once changes are made, we can merge it in development to try it out there.

roles/netbootxyz/templates/menu/k3os.ipxe.j2
pvelati: Ok I've added those variables

antonym (approved):

PR #669, opened 2020-07-04
Add an example of using custom menus with the Docker build

jmgilman:
The current documentation under `/etc/netbootxyz/custom` gives an example of adding custom menus if the end-user is running the playbook locally. However, these instructions are not applicable when using the Docker build as the user does not have the ability to modify the filesystem on the Docker container without manually editing the build file. 

This PR adds an example of how to configure the `custom_templates_dir` variable to have the custom menus built when using Docker.


antonym (approved):

antonym:
Thanks!

PR #713, opened 2020-10-06
Fixed url for anarchy linux

zellkou:
Fixed url for anarchy linux for their new domain anarchyinstaller.org.


antonym:
Thanks, looks like they've also transitioned to GitLab ( https://gitlab.com/anarchyinstaller/installer ) from Github as well, so we'll probably need to change our automation to pull new versions from there instead.

antonym (approved):

Issue #720, opened 2020-10-18
Arch Linux removed archiso initrd


mchangrh:
Arch Linux seems to have removed the archiso.img file in favor of initramfs-linux.img

**To Reproduce**
attempt to install Arch Linux from any of the menus (also broken on dev branch)

**Expected behavior**
When attempting to install, it successfully loads vmlinux-linux but returns an error:
`https://mirror-/archlinux/iso/2020.10.01/arch/boot/x86_64/archiso.img... No such file or directory` and then "Could not boot..."

**Additional context**
This seems to be from the newest 2020.10.01 release, as both 2020.09.01 and 2020.08.01 both still have the archiso.img file.

http://mirror.csclub.uwaterloo.ca/archlinux/iso/2020.09.01/arch/boot/x86_64/
http://mirror.csclub.uwaterloo.ca/archlinux/iso/2020.10.01/arch/boot/x86_64/

mchangrh:
seems like dependent OSes aren't affected
 - [x] Anarchy
 - [x] Black Arch
 - [x] Bluestar
 - [x] Velt
 - [x] SystemRescueCD
 - [x] Zen Install Arch

mchangrh:
Closed in #721

PR #721, opened 2020-10-18
updated archiso.img to initramfs-linux.img

mchangrh:
not sure if other OSes using arch as a base will be affected, but this should allow arch to boot if it was just a rename of archiso.img

Couldn't find any new or information from skimming the site


mchangrh:
this should help resolve #720 if other arch-dependent OSes are not impacted

antonym (approved):
Thanks, looks like it’s the initramfs turn for a rename, vmlinuz had its turn a few months ago.

PR #723, opened 2020-10-22
Add Ubuntu 20.10 Groovy Gorilla

antonym:
Adds groovy, drops eoan, updates live menu for new version
and drops 19.10.


carroarmato0:
Just when I was looking into how to contribute the addition of 20.10 you already went ahead 15min ago :D Thanks for the awesome work @antonym

Issue #737, opened 2020-10-31
NixOS 20.09 missing netboot artifacts in Hydra


antonym:
It appears that NixOS isn't building netboot artifacts from Hydra anymore for 20.09.  If there's not a location upstream anymore to pull the netboot files, we may have to go build them with docker and release it ourselves.

https://nixos.org/manual/nixos/stable/#sec-booting-from-pxe

An example of a builder I hacked on a while ago:

https://github.com/antonym/nixos-builder/blob/master/.circleci/config.yml

Mic92:
Do you depend on circle CI for building? There is also very good github actions support for nix: https://github.com/cachix/install-nix-action and we have similar tooling for https://github.com/nix-community/nix-travis-ci

antonym:
We're using travis mainly, but for the last release we pulled the netboot builds directly from Hydra as downloads, it appears the netboot jobs are not producing the build artifiacts anymore, so if we can't get those anymore we may just need to build them ourselves.

We previously pulled artifacts from here for last release:
https://hydra.nixos.org/job/nixos/release-20.03/nixos.netboot.x86_64-linux/latest

Mic92:
Ok. It should be straight forward convert your circle ci script to a travis ci job. cc @Lassulus

Mic92:
Where should we push the resulting image?

antonym:
Yeah, I'll go down that route if I need to.  Obviously I'd like to pull the bits directly from NixOS without having to re-host as then they are the most trusted and updated source, so I'd like to understand if there is a new location for the artifacts and if not, we'll see about building the artifacts ourselves and hosting.

Mic92:
I can also checkout why we no longer building the netboot image. We also have some ways to hosting stuff on tarballs.nixos.org.

antonym:
Here's where we pulled the 20.03 artifacts from: 

https://hydra.nixos.org/build/124244129

Mic92:
I think we have people working on that already: https://github.com/NixOS/hydra/pull/827 (cc @ajs124)

Mic92:
still relevant.

Mic92:
still an issue.

antonym:
https://hydra.nixos.org/job/nixos/release-21.05/nixos.netboot.x86_64-linux/latest

Still looks broken and missing artifacts...

fuu0:
Looks like its fixed now, the latest build is successful:
https://hydra.nixos.org/build/145123607

[...]

PR #742, opened 2020-11-02
nixos: remove 19.09 add 20.09

Lassulus:
not sure what the path should be? this seems to be updated by CI?


antonym:
Yeah, it's handled by CI.  It appears the artifacts we used to pull from Hydra are no longer present.  Do you know where the netboot artifacts are now? https://github.com/netbootxyz/netboot.xyz/issues/737

antonym:
Since this is being worked upstream, I'll close this out but put a PR to remove 19.09 since it's eol.

PR #755, opened 2020-11-08
Adds a hook for putting local-vars.ipxe on tftp

antonym:
Allows for the user to set local variables to be loaded
into the netboot.xyz menu when loading.  This allows for
setting things like github_user to have custom menu's
auto populate on start up.

Setting an ipxe file with var settings in local-vars.ipxe
of their tftp server would allow those options to be loaded.


nwpr (commented):
Hey! Thanks for being that quick!
One remark: I would put the vars before the Hostname/MAC things, because then general vars can be overwritten per host / they can be used in the host specific menu too

PR #832, opened 2021-01-22
Update CentOS Releases

amit177:
The PR deletes the releases 8.0, 8.1, and 8.2 which are no longer available on the mirrors, as well as adds 8.3 and 8 latest.


antonym (approved):

antonym:
Thanks!

PR #871, opened 2021-04-09
Add support for ShredOS

antonym:
Adds support for ShredOS, an updated version of DBAN using nwipe.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ## [2.0.36]
+### Added
+- Support for ShredOS
 
 ## [2.0.35]
 ### Added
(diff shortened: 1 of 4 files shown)


Firminator:
Hey Antony,
I was testing booting of ShredOS and it failed and brought me right back to the netbootxyz main menu (both the 32bit and 64bit version). Can someone check if it's just me or if it's happening to others as well.
Also I noticed that the 64bit version is outdated which probably is not the reason why it failed to boot, but updating the endpoint would be a good starting point to troubleshoot :)
A lot has changed since the last time I checked netbootxyz. Everthing seems to be automated and I'm having trouble to follow all the breadcrumbs. How do you go along with updating the ShredOS version to the latest v2020.05.017_x86-64_0.32.003 (@https://github.com/PartialVolume/shredos.x86_64/releases/download/v2020.05.017_x86-64_0.32.003/shredos-2020.05.017_x86-64_0.32.003_20211111.img )? Am I on the right track thinking the endpoints.yaml file needs to be edited? Currently is says:
```
shredos-x86_64:
    path: /asset-mirror/releases/download/v2020.05.012_x86-64_0.31-72b45dbf/
    files:
    - shredos
    os: shredos
    version: v2020.05.012_x86-64_0.31
```
Also I'm wondering what the 72b45dbf string means?

PR #890, opened 2021-04-29
add rescuezilla to menus

thelamer:
Little rusty here, don't have my local templating dev env setup anymore. Hope this builds.


roles/netbootxyz/templates/menu/rescuezilla.ipxe.j2
```diff
@@ -0,0 +1,28 @@
...
+# Rescuezilla Disk imaging
+# https://rescuezilla.com/
+
+:rescuezilla
+set os {{ releases.rescuezilla.name }}
+menu ${os} Live CD
+{% for key, value in endpoints.items() | sort %}
+{% if value.os == "rescuezilla" %}
+item {{ item.os }}_{{ item.version }} ${space} ${os} {{ item.version }}
+{% endfor %}
```
antonym, line 13: prob need endif on both

PR #920, opened 2021-06-29
Temporarily testing upstream TLS patch

antonym:
Generating builds to test:

https://github.com/ipxe/ipxe/pull/116

Will revert in hopes this makes it upstream


nshalman:
What was the result of the test?

Issue #422, opened 2019-12-14
Add support for boot loader customization


antonym:
Add support for console customization where a user can change the background picture:
https://ipxe.org/cmd/console

or leverage their own color scheme:

https://github.com/netbootxyz/netboot.xyz/blob/master/roles/netbootxyz/files/ipxe/local/colour.h

PR #936, opened 2021-07-29
add the user customizeable custom_menu_include for early changes to ipxe

commonism:
I'm looking for an easy way to customize the menu

```
console --x 1024 --y 768
console --picture http://boot.netboot.xyz/logo.png
```

I'd prefer something with jinja2 extends/block inheritance, but I do not see how to integrate in the current structure.
On the other hand - the include does the current job.


antonym:
Sounds like a good idea, I'll think about the implementation a bit this weekend, I may want some flags to enable because the current implementation will always enable it.  This could tie into together with the color scheme customization I was thinking about doing as well:

https://github.com/netbootxyz/netboot.xyz/issues/422

antonym (approved):

PR #937, opened 2021-07-29
Change ArchLinux mirror back to Rackspace

brycied00d:
It appears that the evowise.com mirror is kaput, every request gets 302'd to
another website. According to https://archlinux.org/mirrors/evowise.com/893/
this broke around 22 July, 2021 and ArchLinux have pulled the mirror from their
rotation already.

This reverts commit 65ea52a3412ba3aa5ae3c7d6076ab6646307fb5a.


antonym:
Thanks, fixed with: https://github.com/netbootxyz/netboot.xyz/pull/938

Issue #952, opened 2021-08-05
Add support for proxy DHCP


notepass:
**Is your feature request related to a problem? Please describe.**
While setting up a netboot server, I had to use a DHCP proxy to inject the boot server IP into the network. This had to be done, as my DHCP server has no support for setting the "next-server" option. When now booting the netboot.xyz instance, it tried to access the .ipxe files not from my tftp server, but from the server provided by my root DHCP. As there was no tftp server running, all requests timed out (although the normal menu showed up). To reproduce this, the docker container in the repo "samdbmg/dhcp-netboot.xyz" can be used, as that starts up a dnsmasq dhcp proxy.

**Describe the solution you'd like**
When a DHCP proxy exist, netboot.xyz should detect it and honor it in some way. It is a bit hard for me to say in which way would be the best, but I would think of something like this:
* Check if next-server is set
  * If no, check if proxydhcp/next-server is set. If yes, use that to access .ipxe files
  * If yes, check if proxydhcp/next-server is set. If yes, show a prompt telling the user that the proxy-provided server is used instead of the server provided by to root dhcp. 
    * But also give the user  a prompt to override this (e.g. by pressing e or something) 
    * Additionally, a ipxe file could be accessed on the root dhcp provided server (e.g. proxy-dhcp-vars.ipxe) which could set default vars to override any of this behaviour

**Describe alternatives you've considered**
N/A

**Additional context**
I'm currently trying to implement this in a fork. If you would consider adding this feature, I might be able to implement it or submit a PR for it.

notepass:
Created PR #953  to implement this feature

antonym:
Looks to be functioning now with a few changes.  Going to close this out but feel free to test with the development endpoint and binaries:

https://s3.amazonaws.com/dev.boot.netboot.xyz/cf44496e311bf2bf597856b22a2b1368d961ef76/index.html

PR #953, opened 2021-08-05
Implemented support for proxydhcp variables (disbaled by default)

notepass:
I've implemented the feature requested in #952   
I'm not entirely sure if the type of implementation matches what the project wants. By default netboot.xyz will behave the same as before. Only after a change to "proxy-dhcp-vars.ipxe" is made, the new logic will be enabled.  
The overall logic is as follows:
* Check if a proxydhcp is set
  * If no
    * Set the "next-server" provided by the normal DHCP as tftp server
  * If yes
    * Request proxy-dhcp-vars.ipxe from the next-server defined by the normal DHCP to attempt automatic conflict resolution
    * Check the content of the variable "force_use_root_dhcp_settings"
      * If it is true: Use the next-server provided by the normal DHCP server as tftp server
      * If it is false: Use the next-server provided by the proxy DHCP server as tftp server
      * If it is not set or another value: Prompt the user to press F5 to accept the settings from the proxy server. If no key is pressed in 4 seconds, the server provided by the normal DHCP will be used

This logic could be simplified by removing the additional config file, but than there could be changes for the end user (e.g. the newly added prompt), or the proxydhcp could be used whenever it is detected. The code can be changed to do either of this, I'm completely open for feedback on this.

**Edit:** The changes currently only impact UEFI booting, as BIOS booting doesn't seem to have access to proxydhcp info


roles/netbootxyz/templates/menu/proxy-dhcp-vars.ipxe
```diff
@@ -0,0 +1,3 @@
+# This script is called when the dhcp server and a proxy dhcp server booth set the next-server directive. It can be used to force usage of one of those servers. Default value is to use the proxy supplied tftp server but ask the user if the root server supplied tftp server should be used instead
```
antonym: Should be able to leverage local-vars.ipxe instead of a new file.  I'm adding a sample file in https://github.com/netbootxyz/netboot.xyz/pull/956

I'd add something like this:
```
### set to load configuration from proxydhcp/next-server
# set use_proxydhcp_settings true
```

roles/netbootxyz/templates/disks/netboot.xyz.j2
```diff
@@ -20,14 +20,26 @@ iseq ${site_name} netboot.xyz || echo ${bold}${fg_whi}Powered by ${fg_gre}netboo
...
-isset ${next-server} && iseq ${filename} {{ site_name }}-undionly.kpxe && goto tftpmenu ||
-isset ${next-server} && iseq ${filename} {{ site_name }}.efi && goto tftpmenu ||
-isset ${next-server} && iseq ${filename} {{ site_name }}-snp.efi && goto tftpmenu ||
-isset ${next-server} && iseq ${filename} {{ site_name }}-snponly.efi && goto tftpmenu ||
-isset ${next-server} && iseq ${filename} {{ site_name }}-arm64.efi && goto tftpmenu ||
+isset ${next-server} && isset ${proxydhcp/next-server} && goto choose-tftp || set tftp-server ${next-server} && goto load-custom-ipxe
+
+:choose-tftp
+# Load "proxy settings" from root server
+chain tftp://${next-server}/proxy-dhcp-vars.ipxe || echo Could not load definitions to resolve proxy DHCP conflict
```
antonym: Switch to
```
chain tftp://${tftp-server}/local-vars.ipxe || echo ${VARS_ERR}
```
We should be able to reuse the local-vars.ipxe file and then it'll just pull from proxy once set.

roles/netbootxyz/templates/disks/netboot.xyz.j2
```diff
@@ -20,14 +20,26 @@ iseq ${site_name} netboot.xyz || echo ${bold}${fg_whi}Powered by ${fg_gre}netboo
...
+# Check if the proxy-dhcp-vars script has made any usable command about how to progress with a next-server and a proxy-next-server being set
+isset ${force_use_root_dhcp_settings} && iseq ${force_use_root_dhcp_settings} true && set tftp-server ${next-server} && goto load-custom-ipxe || echo -n
```
antonym: I'd probably set the var to use_proxydhcp_settings instead to shorten it and then if it's present in local-vars.ipxe it would use it.  Should be safe to assume it's false if it's not set as true or not present and use the regular tftp server.

```
isset ${use_proxydhcp_settings} && iseq ${use_proxydhcp_settings} true && set tftp-server ${proxydhcp/next-server} && goto load-custom-ipxe ||
set tftp-server ${next-server} && goto load-custom-ipxe || 
```

roles/netbootxyz/templates/disks/netboot.xyz.j2
```diff
@@ -20,14 +20,26 @@ iseq ${site_name} netboot.xyz || echo ${bold}${fg_whi}Powered by ${fg_gre}netboo
...
+isset ${force_use_root_dhcp_settings} && iseq ${force_use_root_dhcp_settings} false && set tftp-server ${proxydhcp/next-server} && goto load-custom-ipxe || echo -n
+prompt --key 0x107e --timeout 4000 DHCP proxy detected. Press ${bold}F5${boldoff} to apply setting from it && set tftp-server ${proxydhcp/next-server} || set tftp-server ${next-server}
```
antonym: Switch to a character and shorten the output.  (just want to avoid function key complexities)
```
prompt --key p --timeout 4000 DHCP proxy detected. Press ${bold}p${boldoff} to boot from && set tftp-server ${proxydhcp/next-server} || set tftp-server ${next-server}
```

roles/netbootxyz/templates/disks/netboot.xyz.j2
```diff
@@ -20,14 +20,26 @@ iseq ${site_name} netboot.xyz || echo ${bold}${fg_whi}Powered by ${fg_gre}netboo
...
+goto load-custom-ipxe
+
+
```
antonym: extra line

antonym (commented):
Requested a few changes but I think the general idea makes sense.

roles/netbootxyz/templates/menu/proxy-dhcp-vars.ipxe
notepass: Removed addition script file and placed variable in local-vars

roles/netbootxyz/templates/disks/netboot.xyz.j2
notepass: Updated chaining

roles/netbootxyz/templates/disks/netboot.xyz.j2
notepass: I slightly disagree with this. A reason for using a proxy server can be, that the tftp server provided by the root DHCP is wrong or not reachable. Defaulting the variable use_proxydhcp_settings to false would mean the following logic flow in the moment:

1. Client requests local-vars from the tftp server provided by the root DHCP
2. tftp server provided by the root does not answer or does not have the file
3. Variable will be set to false
4. tftp server given by the root dhcp will be used, altought it isn't reachable

So, I would propose the following implmentation:
1. Grab local-vars from root provided tftp
2. Check if the variable use_proxydhcp_settings is set. If so, apply the requested tftp server
3. If the variable use_proxydhcp_settings is not set, prompt the user to press a key
4. If the key is pressed, the proxy provided tftp server will be used
5. If the key is not pressed, the root provided tftp server will be used

This allows for utilizing the proxy provided tftp, even if the root provided one isn't online/existing

roles/netbootxyz/templates/disks/netboot.xyz.j2
notepass: Removed additional line

roles/netbootxyz/templates/local-vars.ipxe.j2
```diff
@@ -3,3 +3,8 @@
 
 ### set to enable enable github custom menu
 #set github_user my_github_username
+
+# If a DHCP proxy server is found and this variable is set to true, the tftp server provided by the DHCP proxy will be used instead of the tftp server provided by the root DHCP
+# Please note that this variable is always read from the local-vars.ipxe onb the tftp server the root DHCP provided
+# If the variable isn't present at runtime, the user will be queried to press a key to boot from the proxy DHCP prtovided ftfp server
+#set use_proxydhcp_settings=true
```
antonym: iPXE doesn't use the equals, but a space instead on the set command.

roles/netbootxyz/templates/disks/netboot.xyz.j2
```diff
@@ -20,14 +20,24 @@ iseq ${site_name} netboot.xyz || echo ${bold}${fg_whi}Powered by ${fg_gre}netboo
...
-isset ${next-server} && iseq ${filename} {{ site_name }}-snp.efi && goto tftpmenu ||
-isset ${next-server} && iseq ${filename} {{ site_name }}-snponly.efi && goto tftpmenu ||
-isset ${next-server} && iseq ${filename} {{ site_name }}-arm64.efi && goto tftpmenu ||
+isset ${next-server} && isset ${proxydhcp/next-server} && goto choose-tftp || set tftp-server ${next-server} && goto load-custom-ipxe
+
+:choose-tftp
+# Load "proxy settings" from root server
+chain tftp://${tftp-server}/local-vars.ipxe || echo ${VARS_ERR}
+# Check if the proxy-dhcp-vars script has made any usable command about how to progress with a next-server and a proxy-next-server being set
+isset ${use_proxydhcp_settings} || prompt --key p --timeout 4000 DHCP proxy detected. Press ${bold}p${boldoff} to boot from && set use_proxydhcp_settings true || set use_proxydhcp_settings false
```
antonym: I want to say that it'll only support pass or fail conditions, so the third || may fail.

roles/netbootxyz/templates/menu/netinfo.ipxe.j2
```diff
@@ -17,5 +17,14 @@ item --gap Next-server:
 item nextserver ${space} ${next-server}
 item --gap Filename:
 item filename ${space} ${netX/filename}
+
+# Proxy DHCP info
+item --gap DHCP proxy present
+iseq ${platform} pcbios && item proxy ${space} Not supported on BIOS || isset ${proxydhcp/next-server} && item proxy ${space} Yes || item proxy ${space} No
```
antonym: Same here, pass or fail condition, the third || condition will fail.

roles/netbootxyz/templates/menu/netinfo.ipxe.j2
```diff
@@ -17,5 +17,14 @@ item --gap Next-server:
...
+iseq ${platform} pcbios || item --gap DHCP proxy next-server
+iseq ${platform} pcbios || item proxy-nextserver ${space} ${proxydhcp/next-server}
+iseq ${platform} pcbios || item --gap Using DHCP proxy next-server
+iseq ${platform} pcbios || iseq ${proxydhcp/next-server} ${tftp-server} && echo Yes || echo No (Using default next-server ${tftp-server})
```
antonym: pass or fail condition, the third || condition will fail.

antonym (changes requested):

roles/netbootxyz/templates/menu/netinfo.ipxe.j2
notepass: I will need to remove this again, as after some testing I found BIOS based systems working correctly. So this will resolve itself luckily :)

roles/netbootxyz/templates/menu/netinfo.ipxe.j2
notepass: As above, the PCBIOS check will be removed

roles/netbootxyz/templates/disks/netboot.xyz.j2
notepass: I've rewrote and tested it. It should be fine with the current commit

roles/netbootxyz/templates/disks/netboot.xyz.j2
```diff
@@ -20,14 +20,25 @@ iseq ${site_name} netboot.xyz || echo ${bold}${fg_whi}Powered by ${fg_gre}netboo
...
-isset ${next-server} && iseq ${filename} {{ site_name }}-snp.efi && goto tftpmenu ||
-isset ${next-server} && iseq ${filename} {{ site_name }}-snponly.efi && goto tftpmenu ||
-isset ${next-server} && iseq ${filename} {{ site_name }}-arm64.efi && goto tftpmenu ||
+isset ${next-server} && isset ${proxydhcp/next-server} && goto choose-tftp || set tftp-server ${next-server} && goto load-custom-ipxe
+
+:choose-tftp
+# Load "proxy settings" from root server
+chain tftp://${next-server}/local-vars.ipxe || echo ${VARS_ERR}
+# Check if the proxy-dhcp-vars script has made any usable command about how to progress with a next-server and a proxy-next-server being set
+isset ${use_proxydhcp_settings} || prompt --key p --timeout 4000 DHCP proxy detected. Press ${bold}p${boldoff} to boot from && set use_proxydhcp_settings true ||
```
antonym: Can probably drop the last ||

roles/netbootxyz/templates/disks/netboot.xyz.j2
```diff
@@ -20,14 +20,25 @@ iseq ${site_name} netboot.xyz || echo ${bold}${fg_whi}Powered by ${fg_gre}netboo
...
-isset ${next-server} && iseq ${filename} {{ site_name }}-snponly.efi && goto tftpmenu ||
-isset ${next-server} && iseq ${filename} {{ site_name }}-arm64.efi && goto tftpmenu ||
+isset ${next-server} && isset ${proxydhcp/next-server} && goto choose-tftp || set tftp-server ${next-server} && goto load-custom-ipxe
+
+:choose-tftp
+# Load "proxy settings" from root server
+chain tftp://${next-server}/local-vars.ipxe || echo ${VARS_ERR}
+# Check if the proxy-dhcp-vars script has made any usable command about how to progress with a next-server and a proxy-next-server being set
+isset ${use_proxydhcp_settings} || prompt --key p --timeout 4000 DHCP proxy detected. Press ${bold}p${boldoff} to boot from && set use_proxydhcp_settings true || 
+isset ${use_proxydhcp_settings} || set use_proxydhcp_settings false &&
```
antonym: Can drop the &&

roles/netbootxyz/templates/menu/netinfo.ipxe.j2
```diff
@@ -17,5 +17,13 @@ item --gap Next-server:
 item nextserver ${space} ${next-server}
 item --gap Filename:
 item filename ${space} ${netX/filename}
+# Proxy DHCP info (Will only be displayed if proxydhcp is set)
+item --gap DHCP proxy present
+isset ${proxydhcp} && item proxy ${space} Yes || item proxy ${space} No
+isset ${proxydhcp} || item --gap DHCP proxy next-server
+isset ${proxydhcp} || item proxy-nextserver ${space} ${proxydhcp/next-server}
+isset ${proxydhcp} || item --gap Using DHCP proxy next-server
+isset ${proxydhcp} || iseq ${proxydhcp/next-server} ${tftp-server} && echo Yes || echo No (Using default next-server ${tftp-server})
```
antonym:
```
isset ${proxydhcp} && iseq ${proxydhcp/next-server} ${tftp-server} && echo Yes || echo No (Using default next-server ${tftp-server})
```

roles/netbootxyz/templates/menu/netinfo.ipxe.j2
```diff
@@ -17,5 +17,13 @@ item --gap Next-server:
 item nextserver ${space} ${next-server}
 item --gap Filename:
 item filename ${space} ${netX/filename}
+# Proxy DHCP info (Will only be displayed if proxydhcp is set)
```
antonym, line 20: Most of those should be:

```
isset ${proxydhcp} && item ... ||

antonym (changes requested):

notepass:
Sorry for the late reply, I was tageled up in tape drive stuff (heh).  
I've implemented the requested changes

antonym:
Have finally got around to setting up an environment to test with proxy working, so merging this in as is to build some images with it, test, and tune. I think the menu vars need to be reworked a bit as proxydhcp isn't a var thats set that's set but overall looks good. This should at least get us some builds to test and iterate on.

PR #970, opened 2021-08-27
Some ARM64 fixes

hexchain:
1. Correct the `${arch}` parameter for Fedora and CentOS, as they are called `aarch64` in the URLs.
2. Only allow `changebits` between `x86_64` and `i386` and ignore other values, to prevent problems on non-x86 architectures.

Tested on an ARM64 VM.

### DIFF
--- a/roles/netbootxyz/templates/menu/centos.ipxe.j2
+++ b/roles/netbootxyz/templates/menu/centos.ipxe.j2
@@ -10,21 +10,23 @@ goto ${menu} ||
 
 :centos
 clear osversion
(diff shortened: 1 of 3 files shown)


antonym:
Thanks for this PR, I'm taking it a step further and porting some of these changes to a lot of other distress and doing some overdue cleanup on this one, would love to have some testing once it's completed:

https://github.com/netbootxyz/netboot.xyz/pull/971

Going to close out this one for now, but definitely appreciate the suggestions... I want to make it more generic and consistent across all of the various files.

PR #971, opened 2021-08-28
Architecture type refactor to enable arm64 and i386 builds

antonym:
For those operating systems that support arm64 (and i386), this
is a start of a refactor to enable operating systems that have
those capabilities.

This is a work in progress and consists of some overdue cleanup.


antonym:
Merging this first run to see how much works or breaks.

PR #990, opened 2021-09-09
Fix talos arch files

antonym:


ruant:
@antonym Not sure if it's me or what, but this `${arch}` stuff isn't working for me. Is this supposed to be auto populated by netboot.xyz? Or should it be passed in from some "upstream chain"?

PR #1005, opened 2021-09-25
Update alpinelinux.ipxe.j2

msl4437:
A note about UEFI
If you are booting a uefi system you will need to append initrd=initrdname to the kernel options to boot correctly.

https://wiki.alpinelinux.org/wiki/PXE_boot


antonym:
Was something not working?  {{ kernel_params }} provides initrd=initrd.magic which handles that logic.

https://github.com/netbootxyz/netboot.xyz/blob/development/roles/netbootxyz/defaults/main.yml#L99

PR #1026, opened 2021-10-03
 Fix k3os Live bug with wrong arch

### LINKED ISSUE #1002: k3os boot has a bug
### Discussed in https://github.com/netbootxyz/netboot.xyz/discussions/1001

<div type='discussions-op-text'>

<sup>Originally posted by **jwells38654** September 17, 2021</sup>
Can someone fix the netboot.xyz bootstrap file for TFTP to now append /latest/download/k3os.... as it is no longer the correct GitHub URL path to download the K3os images.  Or how can I fix this?  

FYI- Setting the mirror path still appends the incorrect path mention above.</div>

icepic0:
applied Fix #1002 to Live boot from bug #1001.


antonym (approved):
Thanks! Missed that one!

PR #1036, opened 2021-10-14
Add Padded floppy images for some (iLO/iDRAC) Remote Management Consoles

gratuxri:
Hello, 

some new versions of redfish need exactly 1,44mb, so that image can be uploaded. ipxe has already padded version of this images. I think it's helpfull to some people to have this just for download.

Best Regards,
Juri


antonym:
Looks good, I may consider just dropping the other image in the future to avoid confusion too. Thanks for the contrib!

PR #1053, opened 2021-11-16
Adjust ShredOS in README.md (kernel to USB IMG)

Firminator:
ShredOS provides an USB .IMG or an .ISO
I think this was copied over from DBAN and not adjusted. Since this is not kernel I would think the current endpoints.yml entry would need to be adjusted as well.

It currently says
```
shredos-x86_64:
    path: /asset-mirror/releases/download/v2020.05.017_x86-64_0.32.003-72b45dbf/
    files:
    - shredos
    os: shredos
    version: v2020.05.017_x86-64_0.32.003
    arch: x86_64

```
but you would think it should be something like
```
 files:
 - shredos.IMG
```


antonym:
It's extracted from the image as a kernel without the extension, not seeing an issue here so closing.

PR #1072, opened 2022-02-07
Fix Arch Linux when using static IP

### LINKED ISSUE #1057: Arch Linux 2021.12.01 - No such file or directory
https://boot.netboot.xyz/ipxe/netboot.xyz.lkrn

null-dev:
Fixes: https://github.com/netbootxyz/netboot.xyz/issues/1057

### DIFF
--- a/roles/netbootxyz/templates/menu/archlinux.ipxe.j2
+++ b/roles/netbootxyz/templates/menu/archlinux.ipxe.j2
@@ -3,11 +3,11 @@
 # Arch Linux Operating System
 # http://www.archlinux.org
 
-isset ${dhcp-server} || goto static_ip
-set ipparam BOOTIF=${netX/mac} ip=dhcp
 set os_arch ${arch}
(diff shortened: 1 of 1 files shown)


null-dev:
Also maybe it's better to use http://mirrors.edge.kernel.org/ as the Arch mirror instead of http://mirrors.kernel.org/? The latter just redirects to mirrors.edge.kernel.org on some files. See:
image

Normally this isn't a big deal but because archlinux cannot use DNS when booted with a static IP, it is not able to resolve redirects when it fetches it's airootfs. So even with my fix in this PR, it is still not able to boot properly until I change the mirror to http://mirrors.edge.kernel.org/:

image

antonym (approved):
Thanks for the submission, great catch... I'll also look at the mirror situation too.

antonym:
https://github.com/netbootxyz/netboot.xyz/commit/7ff2dd1b810e4406675ab2fb9eba1fc09aa3e342 Switched to mirrors.edge.kernel.org for archlinux

PR #1076, opened 2022-02-11
Change from Packet to Equinix Metal

dustinmiller:
Not sure if I missed anything, especially in netboot.xyz-metal.j2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+## [2.0.56]
+#### Fixed
+- Change from Packet to Equinix Metal
+
 ## [2.0.55]
 #### Fixed
 - Hardset Flatcar Linux initrd to fix booting (https://github.com/netbootxyz/netboot.xyz/issues/1070)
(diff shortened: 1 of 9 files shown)


nshalman:
Perhaps add a line to `CHANGELOG.md`?

dustinmiller:
> Perhaps add a line to `CHANGELOG.md`?
Definitely, added for us, thanks :smiley:

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+## [2.0.56]
```
nshalman, line 7: Might want to leave it in the Unreleased block and let the maintainers update a release version separately...

antonym: No worries, I'll fix before release.

antonym (approved):
Thanks, was going to switch this previously but figured it may break things on your side. I'll slip this in to the release this weekend.

mmlb:
Actually I am a little worried about this. We set that for everything not just for netboot.xyz so if we do go with tinkerbell/boots#236 we can potentially be breaking other users. I'd rather have metal be an alias for packet, but then again idk if we'd ever actually change it so what would be the point...?

PR #1080, opened 2022-02-24
revert: Equinix Metal is still Packet

dustinmiller:
The previous commit cause some trouble with our ability to access SOL, lets revert for now since our boot server boots hasn't been updated to provide the new ipxe_cloud_config value and thus necessary config isn't being picked up on.

### DIFF
--- a/roles/netbootxyz/defaults/main.yml
+++ b/roles/netbootxyz/defaults/main.yml
@@ -475,8 +475,8 @@ releases:
       name: Digital Ocean
     - key: gcp
       name: GCP
-    - key: metal
-      name: Equinix Metal
+    - key: packet
+      name: Packet
     - key: vmware
       name: VMware
   tinycore:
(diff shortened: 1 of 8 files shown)


antonym:
Would it be possible to reference the existing release packet files instead of reverting?

https://github.com/netbootxyz/netboot.xyz/releases/tag/2.0.54

antonym:
I could also add this to the boot.cfg for now:
```
iseq ${ipxe_cloud_config} metal && goto metal ||
iseq ${ipxe_cloud_config} packet && goto metal ||
```

That would ensure that the SOL settings are getting set correctly.

mmlb:
> Would it be possible to reference the existing release packet files instead of reverting?
> 
> [`2.0.54` (release)](https://github.com/netbootxyz/netboot.xyz/releases/tag/2.0.54)

Not really, a client would usually enter https://boot.netboot.xyz. We don't provide a selectable option/drop down of any sort.

> I could also add this to the boot.cfg for now:
> 
> ```
> iseq ${ipxe_cloud_config} metal && goto metal ||
> iseq ${ipxe_cloud_config} packet && goto metal ||
> ```
> 
> That would ensure that the SOL settings are getting set correctly.

This would work too. We were going to follow up with another PR that was effectively a mix of changing all the strings to reference equinix metal but switch off of `ipxe_cloue_config=packet` as I don't really think we'll ever change it to metal (to avoid breaking other users).

antonym:
Yeah, pointing to the release won't help since you typically us rolling. I'll just go ahead and set:

```
iseq ${ipxe_cloud_config} metal && goto metal ||
iseq ${ipxe_cloud_config} packet && goto metal ||
```

and that should correct things for now to at least align things from this side with the naming and docs. We can always remove the extra check in the future if you do end up cutting over to it.

antonym:
https://github.com/netbootxyz/netboot.xyz/commit/122551012703fddbd1719a6c5fb0c1b5a0838331

mmlb:
sounds good, thanks!

antonym:
I'll try and cut a quick release this weekend so that it'll start working on the rolling release this weekend.

PR #1084, opened 2022-02-27
add pbs

gratuxri:


antonym:
Thanks for the suggestion, refactored a bit and added it here: https://github.com/netbootxyz/netboot.xyz/commit/f3f8e19b0ca4987d538393fb9716ce8896a54497

gratuxri:
Maybe it make sense to include Proxmox Mail Gateway (pmg) as well ;)

antonym:
Yeah, I may do that next... I have not used it before but if the installer method is the same way, it should be easy enough to add.

gratuxri:
installer method is the same

PR #1096, opened 2022-03-16
Use bootloader_filename instead of site_name for bootloader filenames consistently

### LINKED ISSUE #1095: Inconsistent use of bootloader_filename and site_name
The templates have a setting for `bootloader_filename`, but in most places `site_name` is used instead. This causes some features to break if `site_name` and `bootloader_filename` aren't set to the same thing.

For example the `tftpmenu` items aren't loaded and the URLs on the index.html page doesn't work.

abh:
Fixes #1095

### DIFF
--- a/roles/netbootxyz/templates/disks/netboot.xyz.j2
+++ b/roles/netbootxyz/templates/disks/netboot.xyz.j2
@@ -36,12 +36,12 @@ iseq ${use_proxydhcp_settings} true && set tftp-server ${proxydhcp/next-server} 
 goto load-custom-ipxe
 
 :load-custom-ipxe
-isset ${tftp-server} && iseq ${filename} {{ site_name }}.kpxe && goto tftpmenu ||
-isset ${tftp-server} && iseq ${filename} {{ site_name }}-undionly.kpxe && goto tftpmenu ||
-isset ${tftp-server} && iseq ${filename} {{ site_name }}.efi && goto tftpmenu ||
-isset ${tftp-server} && iseq ${filename} {{ site_name }}-snp.efi && goto tftpmenu ||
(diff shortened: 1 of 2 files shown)


abh:
I didn't change the template for the checksum file as that using site_name seemed like it might be intentionally different.

For these files the filename generated was already using `bootloader_filename` and the references I changed were to match the actual files.

abh:
Uh, not sure what's expected here. Bad bot?

abh:
ping @antonym

guppy0130:
I've tried this patch and the following is an example diff in the outputted `index.html`:

```diff
         <tr>
        <td> DHCP-snponly </td>
-       <td> <a href="ipxe/netboot.chicago.inthemainfra.me-snponly.efi">netboot.chicago.inthemainfra.me-snponly.efi</a> </td>
+       <td> <a href="ipxe/netboot.xyz-snponly.efi">netboot.xyz-snponly.efi</a> </td>
        <td> EFI w/ Simple Network Protocol, only boots from device chained from </td>
     </tr>
```

which matches the contents of the `ipxe/` directory correctly:

```console
(netboot.xyz) nyang@omada:~/src/netboot.xyz$ ls buildout/ipxe/
netboot.chicago.inthemainfra.me-sha256-checksums.txt  netboot.xyz.dsk  netboot.xyz.efi  netboot.xyz.kpxe  netboot.xyz.lkrn  netboot.xyz.pdsk  netboot.xyz-snp.efi  netboot.xyz-snponly.efi  netboot.xyz-undionly.kpxe
```

where `netboot.chicago.inthemainfra.me` is the `site_name`. I think this PR is great as-is if the goal is to have valid links/references. could this get merged @antonym ?

antonym (approved):

abh:
thanks @antonym :-)

PR #1108, opened 2022-04-16
add fai

gratuxri:
fai netboot from https://fai-project.org/netboot/


Mrfai:
I'm the author of FAI and it would be nice to add this.

PR #1134, opened 2022-07-14
Update main.yml

apharp:
1. Older releases were moved to a new path.
2. 8.6 was added


antonym:
Thanks, added here: https://github.com/netbootxyz/netboot.xyz/commit/5c07dbc79f54bc8f53f4f5ae50916b6f4863216f

PR #1141, opened 2022-08-05
Adds Oracle 9 and aarch64 support

antonym:


roles/netbootxyz/defaults/main.yml
```diff
@@ -384,8 +384,9 @@ releases:
     mirror: https://yum.oracle.com
     name: Oracle Linux
     paths:
-      7: /repo/OracleLinux/OL7/latest/x86_64
-      8: /repo/OracleLinux/OL8/baseos/latest/x86_64
+      7: /repo/OracleLinux/OL7/latest
+      8: /repo/OracleLinux/OL8/baseos/latest
+      9: /repo/OracleLinux/OL8/baseos/latest
```
chris18890, line 389: should this not be OL9?

antonym: Yeah, great catch, fixed! That's what I get for hacking too late at night. :)

Issue #583, opened 2020-03-08
CentOS Kickstart doesn’t seem to be working


teddue:
After setting the kickstart URL my kickstart file gets ignored in graphical and text mode

hkthomas:
Yes, I also encountered this problem, I tried 3 different installation methods, all failed to install

antonym:
I think https://github.com/netbootxyz/netboot.xyz/blob/development/roles/netbootxyz/templates/menu/centos.ipxe.j2#L51 may need to be inst.ks= instead.  I'll make that change and we can try it out.

PR #1158, opened 2022-09-16
Update fedora kickstart to use inst.ks

bluecmd:
CentOS was fixed back in #583 - add the fix to Fedora as well.

`ks=` is not considered a valid kernel option anymore for Fedora.


antonym (approved):
LGTM, thanks!

PR #1230, opened 2023-04-11
Update VyOS version name from 1.3-rolling to 1.4 rolling

b-:
Netboot.xyz has actually been booting VyOS 1.4-rolling from this menu item for a while now, as it’s the publicly available version. Cosmetically, though, it’s said “1.3”, which can be a little confusing. So this PR fixes that.


antonym (approved):

antonym:
Thanks!

systemmonkey42:
Just a note, the menu now says 1.4, however the version booted is 1.5-rolling.  
This may need to be tweaked again.

PR #1238, opened 2023-04-26
alpine 3.18 -> 3.17.3

brandonros:
I'm pretty sure this will get overriden by automated CI/CD but I couldn't find any evidence that v3.18 exists yet (and when I try to use it, it throws an error about not existing)

### DIFF
--- a/roles/netbootxyz/defaults/main.yml
+++ b/roles/netbootxyz/defaults/main.yml
@@ -131,8 +131,8 @@ releases:
     mirror: http://dl-cdn.alpinelinux.org
     name: Alpine Linux
     versions:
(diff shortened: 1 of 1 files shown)


antonym:
If you can change it to 3.17 I can pause CI. I think it keeps keying off some early releasing. 

https://github.com/netbootxyz/netboot.xyz/commit/d74f1db1cd427d9263a5d147a91b2ca56e65daa7

antonym:
Reverted: https://github.com/netbootxyz/netboot.xyz/commit/d2751e6a3b450935367cd714cb997d38f7a932f6

PR #1249, opened 2023-05-29
Began adding the return to previous menu prompt to menues.

n3gwg:
In as much as I use ProxMox Virtual Environment and use netboot.xyz there within, I use the noVNC console. Hitting the escape key takes the noVNC console out of full screen mode. Thus, I saw the need to add a return to previous menu function to the menues. More menues will likely get this feature added as time goes forth.


n3gwg:
Saw a few mistakes. Will put in a new pull request.

PR #1250, opened 2023-05-29
Add return to previous menu option

n3gwg:
First round of changes to add this functionality, though other menues will be updated as well in another PR.


n3gwg:
I will rebase the commits and resubmit the PR.

PR #1251, opened 2023-05-29
Began adding the return to previous menu prompt to menues.

n3gwg:
A rebased version of my commits for adding return to main menu option to a number of the menus. As I continue to work with the code I will likely have more commits that implement this feature in other menus as well.

### DIFF
--- a/.github/workflows/pull-requests.yml
+++ b/.github/workflows/pull-requests.yml
@@ -21,7 +21,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install ansible ansible-lint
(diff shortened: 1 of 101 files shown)


n3gwg:
I have added more updates. Is updating this pull request the proper way to request their merger into the main code? I hope so. I am not a git expert so just trying to whitestick my way through this.

antonym:
I get the idea but it's a bit of overkill in code to put on each menu for a very specific use case instead of using ESC to jump back a menu. 

Have you tried: https://stackoverflow.com/questions/65724408/proxmox-novnc-prevent-remap-the-escape-key

antonym:
Adds a lot of unnecessary code when esc is built in and functional, closing this out.

PR #1252, opened 2023-05-30
Support building images on other architectures

kaechele:
This unconditionally cross compiles all iPXE images. This has the effect that the build will always produce the correct output for a defined architecture regardless of host architecture.

syslinux is x86 architecture dependent. It doesn't seem to be used in the build process, so the packages are removed from the list of required packages.


kaechele:
Uses https://github.com/netbootxyz/docker-builder/pull/32

antonym:
Had to add syslinux back as it uses it to build the Hybrid ISOs currently. Will see if there are other options to work around that.

kaechele:
Thanks. I didn't notice that during my testing because iPXE silently skips this step if `isohybrid` is not available.

Looking into this.

PR #1275, opened 2023-07-25
Update AlmaLinux 8 to 8.8

codyro:
- 8.7 was moved to the vault, so default 8.7 repository links don't work out of the box


antonym:
Covered by this, will look at adding support for hard set versions from vault at a later date.

https://github.com/netbootxyz/netboot.xyz/commit/c9e1279fc9bcc59435ea13cdf082e9eb19833d6b

PR #1282, opened 2023-09-05
nixos: support boot with aarch64 image

5aaee9:
nixos-images has arm64 image since 23.05 release

### DIFF
--- a/roles/netbootxyz/templates/menu/linux-arm.ipxe.j2
+++ b/roles/netbootxyz/templates/menu/linux-arm.ipxe.j2
@@ -12,6 +12,7 @@ item debian ${space} Debian
 item coreos ${space} Fedora CoreOS
 item gentoo ${space} Gentoo
 item fedora ${space} Fedora
+item nixos ${space} NixOS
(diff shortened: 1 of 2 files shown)


roles/netbootxyz/templates/menu/linux-arm.ipxe.j2
```diff
@@ -18,6 +18,7 @@ item rhel ${space} Red Hat Enterprise Linux
 item rockylinux ${space} Rocky Linux
 item ubuntu ${space} Ubuntu
 item photon ${space} VMware Photon
+item nixos ${space} NixOS
```
antonym: Can we move this up to be alphabetical?

antonym (changes requested):
Adjust menu entry

roles/netbootxyz/templates/menu/linux-arm.ipxe.j2
5aaee9: Updated

antonym (approved):

PR #1317, opened 2023-10-31
make Supergrub2 mirror configurable via boot.cfg

lordrasmus:


antonym:
Closing as want to avoid overrides in that file.

PR #1349, opened 2023-11-29
Add arm64 image generation and allow for independent image generation

nickshanks347:
This merge does two things:

1. Allows for independent image generation for ISO/IMG
2. Generate arm64 ISO/IMG files

ISO/IMG are only generated when `generate_disks_legacy`, `generate_disks_arm` and `generate_disks_efi` are set to `true` as well as `generate_disks_hybrid`. I believe any image should be able to be generated independently without having to have all of them generated. The code looks at the variables set and addresses this issue by explicitly declaring dependencies

The second change adds arm64 ISO/IMG generation. With an increasing number of desktops/SBCs using arm64, images for this architecture would be useful.

May be worth adding in separate images for legacy and EFI so image generation times can be cut down.

### DIFF
--- a/roles/netbootxyz/tasks/generate_disks.yml
+++ b/roles/netbootxyz/tasks/generate_disks.yml
@@ -33,7 +33,4 @@
   ansible.builtin.include_tasks: generate_disks_hybrid.yml
   when:
     - generate_disks_hybrid | default(false) | bool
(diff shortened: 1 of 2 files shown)


antonym (approved):
Thanks!

PR #1362, opened 2023-12-08
Support make -j switch.

markusreil:
* Introduce new variable "make_num_jobs" to user_overrides.yml (default value is 1).
* Use new variable in each task that uses "make".


antonym (approved):
LGTM, I'm also going to set that flag into defaults as well. Thanks!

PR #1410, opened 2024-02-04
Fix missing hash in Talos template

gmemstr:
This prevented Talos from booting with a `Exec format error` from iPXE


antonym:
Thanks, great find, must have clipped it out by accident a while back.

antonym (approved):

Issue #1441, opened 2024-03-26
SystemRescue: Alternative mode with "rootpass=MAC_ADDRESS"


lars18th:
**Is your feature request related to a problem? Please describe.**
The System Rescue CD (inside the Tools menu) has the SSH server enabled by default. However, the password is not set. To enable it you can use the `rootpass` boot parameter. The idea is to set this parameter optionally as the value of the mac address.

**Describe the solution you'd like**
The solution is very straightforward, and it's described here: #1440 . We only need to enable a secondary entry in the menu that enables the parameter.

**Describe alternatives you've considered**
The alternative is to use a custom.ipxe menu. However, this seems to be unnecessary as the suggestion is simple, easy to implement, and doesn't break the security because it will add a _new_ entry, that prints the password. Therefore, it will only be enabled if the user wants it.

**Additional context**
I hope the developers agree with this suggestion and will implement it soon.

lars18th:
Hi @antonym (and other developers),

Please could you add this simple enhancement (#1440) ?
Thank you.

antonym:
Implemented

PR #1471, opened 2024-06-04
Add alternative mode for System Rescue with 'rootpass' as MAC address

antonym:
Related to #1441

Adds a new menu entry to the System Rescue CD Tools menu in `roles/netbootxyz/templates/menu/systemrescue.ipxe.j2` to enable setting the 'rootpass' boot parameter to the device's MAC address.

- **New Menu Entry**: Introduces an option labeled "Enable rootpass=MAC_ADDRESS" that allows users to set the 'rootpass' boot parameter to their device's MAC address for easier access.
- **Toggle Functionality**: Implements a toggle mechanism that enables or disables the 'rootpass' parameter based on the user's selection from the menu. This is achieved through a conditional check that sets or unsets the `rootpass_enabled` flag and the `params` variable accordingly.
- **Dynamic Parameter Setting**: Modifies the boot command to dynamically include the `rootpass` parameter with the MAC address value when the option is enabled by the user.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/netbootxyz/netboot.xyz/issues/1441?shareId=fbe09b2b-f200-48de-8916-542859299786).

### DIFF
--- a/roles/netbootxyz/templates/menu/systemrescue.ipxe.j2
+++ b/roles/netbootxyz/templates/menu/systemrescue.ipxe.j2
@@ -5,6 +5,7 @@ goto ${menu} ||
 :live_menu
 set os System Rescue
 set os_arch ${arch}
+isset ${rootpass_enabled} || set rootpass_enabled false
 iseq ${os_arch} x86_64 && set os_arch amd64 ||
 iseq ${os_arch} i386 && set os_arch i686 ||
 menu ${os}
@@ -15,6 +16,7 @@ item --gap ${os} Versions
 iseq ${os_arch} {{ value.arch }} && item {{ value.version }}_${os_arch} ${space} ${os} {{ value.version }} ||
 {% endif %}
 {% endfor %}
+item rootpass_mac Enable rootpass=MAC_ADDRESS [ enabled: ${rootpass_enabled} ]
 choose live_version || goto live_exit
 goto ${live_version}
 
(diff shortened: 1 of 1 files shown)


lars18th:
Hi @antonym ,

Thank you for incorporating my idea suggested in #1440 .

However, the current implementation prints in the screen menu `Enable rootpass=MAC_ADDRESS`. While the idea is to print the current MAC address: `Enable rootpass=00aabbcc1122`. You can check my original dirty implementation in https://github.com/lars18th/netboot.xyz-custom/blob/master/custom.ipxe

This has sense because with multiple NICs the MAC ADDRESS could not be clear.

I hope you will want to improve this.
Regards.

lars18th:
ping @antonym

antonym:
https://github.com/netbootxyz/netboot.xyz/commit/d1ad272087c9afb1077a0ee8a88f22d8e396c64b

lars18th:
Thank you @antonym . 👍 
Waiting for the release with this change.

lars18th:
Hi @antonym ,

After testing it... You've missed to disable the firewall!
Read here about the boot cheatcodes: https://www.system-rescue.org/manual/Network_configuration_and_programs/
In section **Running an SSH Server** it indicates not only to use `rootpass=xxx` but to add `nofirewall` to enable incoming connections.

Please, fix it.
Thank you.

lars18th:
And another missing thing (sorry, in this case it's my fault): I suggest to print (in the menu) the IP address used in the interface with the macaddress to know it. Like:
```
item rootpass_mac Enable rootpass=MAC_ADDRESS (for IP: IP_ADDRESS) [ enabled: ${rootpass_enabled} ]
```
😉

PR #1525, opened 2024-10-15
Add DTS v2.0.0

PLangowski:
Dasharo Tools Suite (DTS) is a set of tools running in a minimal Linux environment to deploy, update, and maintain firmware on Dasharo-supported devices. For example, it can be used to update the firmware on a device or run the initial deployment, even when no OS is currently installed. https://docs.dasharo.com/dasharo-tools-suite/overview/

### DIFF
--- a/endpoints.yml
+++ b/endpoints.yml
@@ -394,6 +394,13 @@ endpoints:
     - vmlinuz
     os: caine
     version: '13.0'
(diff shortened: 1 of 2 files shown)


antonym (approved):
LGTM, will merge into development to test functionality.

PR #1542, opened 2024-11-23
feat(windows): Add support for booting Windows on arm64

ashlcx:
Adds support for booting Windows on arm64 using wimboot.arm64
* add windows to arm64 boot menu
* add wimboot.arm64 to menus
* add arm64 arch in windows arch select
* adds conditional for kernel

### DIFF
--- a/roles/netbootxyz/templates/menu/windows.ipxe.j2
+++ b/roles/netbootxyz/templates/menu/windows.ipxe.j2
@@ -43,7 +43,7 @@ isset ${win_base_url} && goto boot || echo URL not set... && goto url_set
 
 :boot
 imgfree
-iseq ${win_arch} arm64 kernel {{ wimboot_arm_location }} || kernel {{ wimboot_location }}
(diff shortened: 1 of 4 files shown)


roles/netbootxyz/templates/menu/windows.ipxe.j2
```diff
@@ -19,7 +19,7 @@ choose win_version || goto windows_exit
 goto ${win_version}
 
 :arch_set
-iseq ${win_arch} x64 && set win_arch x86 || set win_arch x64
+iseq ${win_arch} x64 && set win_arch x86 && set win_arch arm64 || set win_arch x64
```
antonym, line 22: This logic may not work well, as it's flipping between x86 and arm64, so x86 will probably never get set.

ashlcx: Ah yeah good point, I'll add in automatic arch detection and a seperate menu as well

PR #1558, opened 2025-01-05
Add OpenBSD 7.6 & 7.5, drop EOL versions

job:
While there, use the CDN with HTTPS instead of the (high latency) master site

### DIFF
--- a/roles/netbootxyz/defaults/main.yml
+++ b/roles/netbootxyz/defaults/main.yml
@@ -369,21 +369,18 @@ releases:
     base_dir: pub/OpenBSD
     enabled: true
     menu: bsd
-    mirror: http://ftp.openbsd.org
+    mirror: https://cdn.openbsd.org
     name: OpenBSD
     versions:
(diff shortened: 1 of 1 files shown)


roles/netbootxyz/defaults/main.yml
```diff
@@ -369,21 +369,18 @@ releases:
...
     name: OpenBSD
     versions:
-    - code_name: '7.4'
-      image_ver: '74'
+    - code_name: '7.6'
+      image_ver: '76'
+      name: '7.6'
+    - code_name: '7.5'
+      image_ver: '75'
       name: '7.4'
```
antonym: Switch this to 7.5 as well, but other than that LGTM.

antonym (changes requested):

roles/netbootxyz/defaults/main.yml
job: done!

antonym (approved):

PR #1560, opened 2025-01-13
Sort NixOS releases newest to oldest

mweinelt:
This means the first version shown is the latest release, which makes more sense.

The nixos-unstable release unfortunately hasn't been rebuilt in years, so I'm not sure if it makes sense to include it. cc @mic92


antonym:
Updated the bot:

https://github.com/netbootxyz/netboot.xyz/commit/3155a000c4275e8b6bdeed8cf5bba97c5bb45ecd

PR #1625, opened 2025-05-17
feat: Add Flatcar ARM64 support

nalum:
I think this is correct based on other templates but not sure on how best to test it.

Any pointers on testing?

### DIFF
--- a/roles/netbootxyz/templates/menu/flatcar.ipxe.j2
+++ b/roles/netbootxyz/templates/menu/flatcar.ipxe.j2
@@ -11,7 +11,10 @@ goto ${menu}
 
 :flatcar
 set os {{ releases.flatcar.name }}
-menu ${os}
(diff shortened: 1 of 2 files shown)


nalum:
Added these changes to my local deployment via the UI and I'm getting the following error when I select it:

```
flatcar.ipxe... ok
{%: command not found
Could not boot: Exec format error (https://ipxe.org/2e022081)
Could not boot: No such file or directory (https://ipxe.org/2d02808e)
Error occurred, press any key to return to menu ...
```

Will dig into this.

Edit: Ah this is the templating, will retry with the templating replaced.

nalum:
That's worked, successfully booted into flatcar arm64 on a compute blade with raspberry pi cm4.

antonym (approved):
LGTM, thanks for the submission!

PR #1638, opened 2025-06-16
fix(ipxe): fix filename checks

### LINKED ISSUE #1097: tftpmenu not loaded if tftp filename has a path component
I load ipxe over tftp from a path (`ipxe/netboot.xyz.kpxe`). If I'm not mistaken (which is very possible) that's what makes this line from `netboot.xyz.j2` not match:

```
isset ${tftp-server} && iseq ${filename} {{ bootloader_filename }}.kpxe && goto tftpmenu ||
```

It's not clear to me what the `iseq` part is supposed to be accomplishing. I am guessing there are some filenames that shouldn't go to `tftpmenu`, but if so I'd suggest those should be called out rather than the opposite. (And if they shouldn't be going to `tftpmenu`, would they ever be running in a context with `tftp-server` set?).

I believe this section should be replaced with just a line like:

    isset ${tftp-server} && goto tftpmenu || goto menu

(at least for my use case this fixed it so I can use `MAC-[mac-address].ipxe`).

jkoelker:
The `filename` contains the full path to the file, usually the netboot assets are in `/` of a tftp server, so the check if its booted from the ipxe assets will always fail. Add an additional check prepended with `/` to allow the check to pass when booted from the root of a tftp server.

Also add loading of MAC- ipxe scripts with hypens in addition to raw, i.e. `MAC-00-11-22-33-44-55.ipxe` or `MAC-001122334455.ipxe`.

Fixes: #1097

### DIFF
--- a/roles/netbootxyz/templates/disks/netboot.xyz.j2
+++ b/roles/netbootxyz/templates/disks/netboot.xyz.j2
@@ -15,6 +15,7 @@ set boot_domain {{ boot_domain }}
 set ipxe_version ${version}
 set version {{ boot_version }}
 set conn_type https
+set tftp-root-path /
 
 :start
 echo ${bold}${fg_gre}${site_name} - ${fg_whi}v${version}${boldoff}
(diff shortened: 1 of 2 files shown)


jkoelker:
I deployed a new environment and found the MAC overrides were not working anymore:

image

I don't think the existing `isset` will ever match, but left it in case someone is relying on it.

roles/netbootxyz/templates/disks/netboot.xyz.j2
```diff
@@ -98,7 +105,7 @@ goto menu || goto failsafe
 :tftpmenu
 chain tftp://${tftp-server}/local-vars.ipxe || echo ${VARS_ERR}
 isset ${hostname} && chain --autofree tftp://${tftp-server}/HOSTNAME-${hostname}.ipxe || echo Custom boot by Hostname not found trying MAC...
-chain --autofree tftp://${tftp-server}/MAC-${mac:hexraw}.ipxe || echo Custom boot by MAC not found booting default...
+chain --autofree tftp://${tftp-server}/MAC-${mac:hexraw}.ipxe || chain --autofree tftp://${tftp-server}/MAC-${mac:hexhyp}.ipxe || echo Custom boot by MAC not found booting default...
```
antonym: I think a third check will break on one line, maybe switch to:

```
isset ${hostname} && chain --autofree tftp://${tftp-server}/HOSTNAME-${hostname}.ipxe || echo Custom boot by Hostname not found trying MAC (hexraw)...
chain --autofree tftp://${tftp-server}/MAC-${mac:hexraw}.ipxe || echo Custom boot by MAC (hexraw) not found, attempting by MAC (hexhyp)...
chain --autofree tftp://${tftp-server}/MAC-${mac:hexhyp}.ipxe || echo Custom boot by MAC (hexhyp) not found, booting default...
```

antonym (changes requested):

roles/netbootxyz/templates/disks/netboot.xyz.j2
jkoelker: Totes, updated!

antonym:
@claude

antonym:
@claude

antonym:
@claude

PR #1640, opened 2025-06-19
alpinelinux: use interface autodetection

clandmeter:
When there are multiple network cards in a system, eth0 could possible not be the right interface to configure leaving the system unable to boot. Leaving it empty will try to find the first interface which is active. This should also provide a better error message when no active interface is found instead of just exiting into an emergency shell.


antonym:
@claude review

PR #1644, opened 2025-06-29
feat(talos): add overridable timeout for auto install

ruant:
I guess most people just want talos to get autoinstalled. I do atleast.

I set the timeout to be 60 seconds by default, but perhaps a better solution is to keep todays functionality, aka do not proceed and someone has to manually hit enter. Not sure how that is done tho?


antonym:
@claude review

antonym:
@claude please review

PR #1652, opened 2025-08-04
Create custom EFI boot for Rampart-AIOS

ginanjar-rahmansyah:
Changes

- Update default netboot.xyz.j2 to only trigger our custom ipxe script
- To reduce build duration from 20 min to 10 min by only build EFI boot


jkoelker (changes requested):
I think this is a pr by mistake. We don't want to point everything at another source.

ginanjar-rahmansyah:
My bad, let me remove it

PR #1687, opened 2025-10-14
Add ObsidianOS

TheOddCell:
https://obsidianos.xyz/
I am only adding the base image as the other images (KDE and Cosmic) are graphical and are unreasonably big.

### DIFF
--- a/roles/netbootxyz/templates/menu/obsidianos.ipxe.j2
+++ b/roles/netbootxyz/templates/menu/obsidianos.ipxe.j2
@@ -1,0 +1,14 @@
+#!ipxe
+
+# ObsidianOS
+# https://obsidianos.xyz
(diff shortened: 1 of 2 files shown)


neoapps-dev:
fire

roles/netbootxyz/templates/menu/obsidianos.ipxe.j2
```diff
@@ -0,0 +1,14 @@
+#!ipxe
+
+# ObsidianOS
+# https://obsidianos.xyz
+
+imgfree
+kernel https://files.obsidianos.xyz/latest.iso || goto failed
```
antonym, line 7: Not valid, kernel command can't load an ISO.

PR #1722, opened 2025-12-14
Update OpenBSD, enable BSD on UEFI

klemensn:
Built on Ubuntu 22.04 and boot tested on real hardware as well as in QEMU.

The only currently supported BSDs are FreeBSD and OpenBSD.

FreeBSD overall remains unchanged, i.e. still only available on amd64+pcbios.

OpenBSD becomes available on arm64/amd64+efi.

### DIFF
--- a/roles/netbootxyz/defaults/main.yml
+++ b/roles/netbootxyz/defaults/main.yml
@@ -401,18 +401,18 @@ releases:
     base_dir: pub/OpenBSD
     enabled: true
     menu: bsd
-    mirror: http://ftp.openbsd.org
+    mirror: http://cdn.openbsd.org
     name: OpenBSD
     versions:
+    - code_name: snapshots
+      image_ver: '78'
+      name: 7.8 Latest Snapshot
+    - code_name: '7.8'
+      image_ver: '78'
+      name: '7.8'
     - code_name: '7.7'
       image_ver: '77'
       name: '7.7'
-    - code_name: '7.6'
-      image_ver: '76'
-      name: '7.6'
-    - code_name: snapshots
-      image_ver: '77'
-      name: 7.7 Latest Snapshot
   opensuse:
     base_dir: distribution/leap
     enabled: true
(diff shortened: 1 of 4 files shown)


antonym:
Memdisk will only work under legacy, not UEFI or Arm unfortunately.

klemensn:
> Memdisk will only work under legacy, not UEFI or Arm unfortunately.

Correct.

Is this just a remark or do you mean to point out an error I made?

antonym:
Mainly that the arm and efi menus were toggled on, and probably won't work.

klemensn:
> Mainly that the arm and efi menus were toggled on, and probably won't work.

Like I wrote in the PR's description, I did test this on both arm64 and amd64, both UEFI and BIOS, of course:

- amd64
  - UEFI
    - OpenBSD boots
    - FreeBSD is not shown
  - BIOS
    - OpenBSD boots
    - FreeBSD boots
- arm64
  - UEFI
    - OpenBSD boots
    - FreeBSD is not shown

I now retested again, including full OpenBSD installations to disk via this PR, and see the same results:

- amd64/BIOS remains working as before
- amd64/UEFI gains OpenBSD
- arm64/UEFI gains OpenBSD

See commit messages for more details.

Feel free to ask about specifics here if anything is still unclear.

antonym (approved):

klemensn:
I just booted from https://boot.netboot.xyz and noticed this PR being only partially applied:
- cdn.openbsd.org is used instead of ftp.o.o
- 7.8 is available, 7.6 is gone

Checking /{menu,bsd,openbsd}.ipxe and /boot.cfg the above mentioned bits from this PR indeed landed,
whereas the rest did not:
- installXY still used over minirootXY
- no BSD menu on UEFI

I tried to find information on when and how boot.n.x gets updated, but could not find any.

@antonym Could you help explain how we got this inconsistent state?

antonym:
If you want to use it immediately, you can do so via the Utility Menu and select development endpoints. Version numbers usually increment out to the rolling release on boot.netboot.xyz immediately but menu alterations and changes are coupled with the main releases so it'll will fully be pushed out once a new release is cut. The version numbers for the OS distroos increment automatically so that the menus can still be fresh between full releases.

PR #1798, opened 2026-05-20
update pull

vampiricbunny:
made it so user can just drop new iso in the iso folder (or in folders in the iso folders


antonym:
Memdisk unreliable for ISO, doesn't work with UEFI

vampiricbunny:
i tested it on uefi worked fine ty for reply tho basically i just want a easy way to drop a iso in there current setup is harder but does basic same thing from what i have seen in docker (with or with out uefi )

Sent with [Proton Mail](https://proton.me/mail/home) secure email.

